PR #5, opened 2018-03-29
Rdm 1925 Check Your Answers page is not showing all the information about additional executors

andrewfolga:

```
[ ] Yes
[x ] No
```

### DIFF
--- a/src/app/shared/palette/complex/read-complex-field.component.ts
+++ b/src/app/shared/palette/complex/read-complex-field.component.ts
@@ -8,21 +8,4 @@ import { FieldsUtils } from '../../utils/fields.utils';
   templateUrl: './read-complex-field.html',
   styleUrls: ['./read-complex-field.scss']
 })
-export class ReadComplexFieldComponent extends AbstractFieldReadComponent {
-
-  getComplexFields() {
-    let complexFields = [];
-
-    this.caseField.field_type.complex_fields.forEach((elem, index) => {
-      let elemCopy = FieldsUtils.cloneObject(elem);
-      if (this.caseField.value) {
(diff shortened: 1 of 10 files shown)


src/app/shared/palette/complex/read-complex-field.component.ts
```diff
@@ -1,9 +1,30 @@
...
+        elemCopy.value = this.caseField.value[elem.id];
+        if (!elem.value) {
+          complexFields[index] = elemCopy;
+        }
+      }
+    });
+    return complexFields;
+  }
+
+  private cloneObject(obj: any): any {
```
mario-paniccia: we have in field.utils.ts the same method. At the moment is private. Shall we make it public and reuse that one?

src/app/shared/palette/complex/read-complex-field.component.ts
```diff
@@ -1,9 +1,30 @@
...
 
 @Component({
   selector: 'ccd-read-complex-type-field',
   templateUrl: './read-complex-field.html',
   styleUrls: ['./read-complex-field.scss']
 })
-export class ReadComplexFieldComponent extends AbstractFieldReadComponent {}
+export class ReadComplexFieldComponent extends AbstractFieldReadComponent {
+
+  getComplexFields() {
```
mario-paniccia: not sure if this is an issue but if we make a copy of the complex fields, and the from the ui we update the copied fields, does the parent type sees the update?

andrewfolga:
There is some tech debt introduced earlier about the way we merge case fields and form fields for later use in label substitution and conditional show hide directives that will be addressed in https://tools.hmcts.net/jira/browse/RDM-1980.

PR #29, opened 2018-04-13
add-vault-to-project

agilks:
### Change description ###

```
The change is to introduce  way for the project to access values important /
 to the execution of automated scripts, in a way that these values are not /
exposed to the public or reference directly by the project code /

A method has been added to main.tf which enables the retrieval /
of those secrets which can then be set as environment variables  /
and supporting varibales added to variables.tf

The smoke tests are now disable to enabled the project to build on CNP 

```

```
[ ] Yes
[x] No
```

### DIFF
--- a/infrastructure/main.tf
+++ b/infrastructure/main.tf
@@ -35,12 +35,12 @@ module "case-management-web" {
 
   module "ccd-case-management-web-vault" {
     source              = "git@github.com:hmcts/moj-module-key-vault?ref=master"
-    name                = "ccd-case-management-web${var.env}" // Max 24 characters
+    name                = "ccd-case-web-${var.env}" // Max 24 characters
     product             = "${var.product}"
     env                 = "${var.env}"
     tenant_id           = "${var.tenant_id}"
     object_id           = "${var.jenkins_AAD_objectId}"
-    resource_group_name = "${module.ccd-case-management-web.resource_group_name}"
+    resource_group_name = "${module.case-management-web.resource_group_name}"
     product_group_object_id = "be8b3850-998a-4a66-8578-da268b8abd6b"
   }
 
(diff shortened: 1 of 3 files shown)


infrastructure/main.tf
```diff
@@ -32,4 +32,16 @@ module "case-management-web" {
     PRINT_SERVICE_URL_REMOTE = "${var.ccd_print_service_url}"
     WEBSITE_NODE_DEFAULT_VERSION = "8.9.4"
   }
+
+  module "ccd-case-management-web-vault" {
+    source              = "git@github.com:hmcts/moj-module-key-vault?ref=master"
+    name                = "ccd-case-management-web${var.env}" // Max 24 characters
```
vlaurin: it needs to be max 24 chars including env name which varies between 3 and 7 chars.
`ccd-case-web-{var.env}` should do the trick

infrastructure/main.tf
```diff
@@ -32,4 +32,16 @@ module "case-management-web" {
...
+    product             = "${var.product}"
+    env                 = "${var.env}"
+    tenant_id           = "${var.tenant_id}"
```
vlaurin: This needs to be added to `variables.tf` (see https://github.com/hmcts/ccd-data-store-api/blob/master/infrastructure/variables.tf#L24)

infrastructure/main.tf
```diff
@@ -32,4 +32,16 @@ module "case-management-web" {
...
+    object_id           = "${var.jenkins_AAD_objectId}"
```
vlaurin: This needs to be added to `variables.tf` (see https://github.com/hmcts/ccd-data-store-api/blob/master/infrastructure/variables.tf#L32)

infrastructure/main.tf
```diff
@@ -32,4 +32,16 @@ module "case-management-web" {
...
+    resource_group_name = "${module.ccd-case-management-web.resource_group_name}"
```
vlaurin: This should be a reference to https://github.com/hmcts/ccd-case-management-web/pull/29/files#diff-1031eeab417cd8afe77e7e005717d38dR9 so:
```
resource_group_name = "${module.case-management-web.resource_group_name}"
```

PR #35, opened 2018-04-13
Set jenkins environment variables for tests

agilks:
### Change description ###

```
This change is to instruct Jenkins to set some environment variables used by the smoke tests. 
```

```
[ ] Yes
[x] No
```

### DIFF
--- a/Jenkinsfile_CNP
+++ b/Jenkinsfile_CNP
@@ -9,6 +9,21 @@ properties(
 def product = "ccd"
 def app = "case-management-web"
 
+List<LinkedHashMap<String, Object>> secrets = [
+    secret('caseworkerAutoTestEmail', 'CCD_CASEWORKER_AUTOTEST_EMAIL'),
+    secret('caseworkerAutoTestPassword', 'CCD_CASEWORKER_AUTOTEST_PASSWORD'),
+    secret('testFrontEndUrl', 'TEST_FRONTEND_URL')
+]
+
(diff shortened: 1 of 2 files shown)


Jenkinsfile_CNP
```diff
@@ -9,6 +9,21 @@ properties(
 def product = "ccd"
 def app = "case-management-web"
 
+List<LinkedHashMap<String, Object>> secrets = [
+    secret('caseworkerAutoTestEmail', 'CCD_CASEWORKER_AUTOTEST_EMAIL'),
+    secret('caseworkerAutoTestPassword', 'CCD_CASEWORKER_AUTOTEST_PASSWORD'),
+    secret('testFrontEndUrl', 'TEST_FRONTEND_URL'),
```
vlaurin: I don't think this one is required. I believe Jenkins already provides a `TEST_URL` environment variable which points to the instance of the app which was just deployed.

Jenkinsfile_CNP
```diff
@@ -9,6 +9,21 @@ properties(
...
+static LinkedHashMap<String, Object> secret(String secretName, String envVar) {
+    [ $class: 'AzureKeyVaultSecret',
+      secretType: 'Secret',
+      name: secretName,
+      version: '',
+      envVariable: envVar
+    ]
+}
+
 withPipeline("angular", product, app) {
```
vlaurin: To load the secrets, the following is required inside `withPipeline`: 

```
loadVaultSecrets(secrets)
```

PR #60, opened 2018-04-25
RDM-1591 Added conditional hide/show for case view

tzarsmango:
https://tools.hmcts.net/jira/browse/RDM-1591

This one is adding conditional hide/show for the CaseView Fields.

The case can be tested using the `stubs-unit` application

```
[ ] Yes
[X] No
```

### DIFF
--- a/src/app/shared/conditional-show/conditional-show-view.directive.ts
+++ b/src/app/shared/conditional-show/conditional-show-view.directive.ts
@@ -1,0 +1,35 @@
+import { AfterViewInit, Directive, ElementRef, Input } from '@angular/core';
+import { CaseField } from '../domain/definition/case-field.model';
+import { ShowCondition } from './conditional-show.model';
+
+@Directive({selector: '[ccdConditionalShowView]'})
+/** Hides and shows the host element based on the show condition if the condition is not empty. Works on read only fields and form fields.
+ *  The show condition is evaluated on all the fields of the page. i.e. read only and form fields. When a form field is hidden, if its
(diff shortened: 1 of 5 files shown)


src/app/shared/conditional-show/conditional-show-view.directive.ts
```diff
@@ -0,0 +1,35 @@
+import { AfterViewInit, Directive, ElementRef, Input } from '@angular/core';
+import { CaseField } from '../domain/definition/case-field.model';
+import { ShowCondition } from './conditional-show.model';
+
+@Directive({selector: '[ccdConditionalShowView]'})
```
mario-paniccia: why a new directive? 
have you tried using  ccdConditionalShow without passing the formGroup property? e.g.
<div ccdConditionalShowView ccdLabelSubstitutor [caseField]="field" [eventFields]="caseFields">

tzarsmango: I started with that option but got into trouble -which I don't remember clearly now- so I decided to remove some logic to have a simple new directive

andrewfolga: Would be nice to reuse the existing directive that @mario-paniccia suggested if it is not too hard to do?

tzarsmango:
The fix for RDM-1591 will be delivered by `RDM-1964`. Will use the existing `ccdConditionalShow` directive as well. So, closing this PR.

PR #64, opened 2018-04-26
Rdm 1964 - Conditional Hide/Show Tabs & TabFields while viewing case

tzarsmango:
https://tools.hmcts.net/jira/browse/RDM-1831

This one is adding conditional hide/show for the CaseViewTabs and Fields on the tabs

The case can be tested using the stubs-unit application

```
[ ] Yes
[X] No
```

### DIFF
--- a/src/app/cases/viewer/case-viewer.component.ts
+++ b/src/app/cases/viewer/case-viewer.component.ts
@@ -12,6 +12,7 @@ import { ActivityService } from '../../core/activity/activity.service';
 import { Observable } from 'rxjs/Observable';
 import { Subscription } from 'rxjs/Subscription';
 import { CaseField } from '../../shared/domain/definition/case-field.model';
+import { ShowCondition } from '../../shared/conditional-show/conditional-show.model';
 
 // TODO make this configurable
 const RETRY = 5;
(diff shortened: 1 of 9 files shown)


src/app/cases/viewer/case-viewer.component.ts
```diff
@@ -43,13 +44,13 @@ export class CaseViewerComponent implements OnInit, OnDestroy {
...
 
-    // Clone and sort fields array
-    this.sortedTabs = this.sortedTabs.map(tab => Object.assign({}, tab, {
-      fields: this.orderService.sort(tab.fields)
-    }));
-
     this.caseFields = this.getTabFields();
 
+    // Clone and sort fields array
+    this.sortedTabs = this.sortedTabs
```
andrewfolga: Can we extract the right side of assignment to a well named method? Maybe the comment could be removed as well.

src/app/cases/viewer/case-viewer.component.ts
```diff
@@ -43,13 +44,13 @@ export class CaseViewerComponent implements OnInit, OnDestroy {
...
+      .map(tab => Object.assign({}, tab, { fields: this.orderService.sort(tab.fields) }))
```
mario-paniccia: We have  a field utils class where we are adding utility method for fields. Things like cloning ecc. Can we reuse the clone method in there or move this 'cloning' method there?

src/app/shared/conditional-show/conditional-show.model.ts
```diff
@@ -30,4 +32,18 @@ export class ShowCondition {
   private removeStarChar(s: string) {
     return s.substring(0, s.length - 1);
   }
+
+  matchByCaseFields(caseFields: CaseField[]): boolean {
```
mario-paniccia: I think this whole method can be moved in FieldsUtils. 
Also, perhaps we could reuse some method already there to achieve the same result

src/app/shared/conditional-show/conditional-show.model.ts
```diff
@@ -30,4 +32,18 @@ export class ShowCondition {
...
+    return this.match(this.buildState(caseFields));
+  }
+
+  private buildState(caseFields: CaseField[]): any {
```
mario-paniccia: I think this whole method can be moved in FieldsUtils. Can we called toValuesMap or similar?
Also, perhaps we could reuse some method already there to achieve the same result

src/app/cases/viewer/case-viewer.component.ts
tzarsmango: 👍

src/app/cases/viewer/case-viewer.component.ts
tzarsmango: Field utils generally has methods for merging caseFields and formFields, but we're having a two level data being merged & sorted for caseTabs and caseTabsFields, so I prefer this one to stay here

PR #74, opened 2018-05-04
RDM-2174

mario-paniccia:
To fix the bug the best approach is to reset invalid form fields when navigating on previous page

```
[ ] Yes
[x ] No
```

### DIFF
--- a/src/app/shared/case-editor/case-edit.component.ts
+++ b/src/app/shared/case-editor/case-edit.component.ts
@@ -85,11 +85,10 @@ export class CaseEditComponent implements OnInit {
 
   previous(currentPageId: string): Promise<boolean> {
     let previousPage = this.wizard.previousPage(currentPageId, this.buildCanShowPredicate());
-
     if (!previousPage) {
       return Promise.resolve(false);
     }
-
+    this.resetInvalidFormFields();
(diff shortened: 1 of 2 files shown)


src/app/shared/case-editor/case-edit.component.ts
```diff
@@ -105,4 +104,13 @@ export class CaseEditComponent implements OnInit {
     this.confirmation = confirmation;
     return this.router.navigate(['confirm'], {relativeTo: this.route});
   }
+
+  private resetInvalidFormFields() {
+    Object.keys(this.form.controls['data'].value).forEach(key => {
+      const control = this.form.controls['data'].get(key);
+      if (control.touched && control.invalid) {
+        control.reset();
```
andrewfolga: Has this been tried for simple, complex and collection types?

mario-paniccia: That's a good point. Haven't tested it for complex tipes yet. Will do cheers

mario-paniccia:
Closing this as Stuart has requested another solution that doesn't require any clearing

mario-paniccia:
Closing this as Stuart has requested another solution that doesn't require any clearing

PR #77, opened 2018-05-08
RDM-1867 - Button behaviour:

MKaramath:
https://tools.hmcts.net/jira/browse/RDM-1867

Updated the button behaviour on the Create Case event:
1.   The button under the 3 dropdowns is changed from Submit to Start
2.   Previous buttons are removed altogether and replaced with a "<Back" link at the top left of the page 
3.   Submit (top right of page) becomes Go; Action to be taken becomes Next step; Widen the event dropdown to max 25 characters.
https://app.zeplin.io/project/5973ca007a07d80043f3b447/screen/5ae85fd0413df387225d64d2

```
[ ] Yes
[x] No
```


src/app/shared/case-editor/case-edit-page.html
```diff
@@ -1,3 +1,4 @@
+<a (click)="previous()" href="javascript:void(0)" class="back">&lt; Back</a>
```
tzarsmango: The control ```[disabled]="!hasPrevious()" ``` seems to be missing here. If there is no previous page, this link should be `hidden` or `disabled`

MKaramath: The Back link is now reverted back to Previous button.   The comment is incorporated.

PR #79, opened 2018-05-09
fixed broken back navigation caused by label component not being able…

mario-paniccia:
The label component is going in exception when a form date field has an invalid date. This breaks the navigation. Fixed that
In case of invalid date, e.g. 13-05, the Label will output:
{ Invalid Date: 13-05 }

Also fixed wrong date format in label. At the moment is displayed as 05-05-1981 but should be 05 May 1981 like we have in the rest of the app

https://tools.hmcts.net/jira/browse/RDM-2174

```
[ ] Yes
[x ] No
```

### DIFF
--- a/src/app/shared/substitutor/label-substitutor.directive.spec.ts
+++ b/src/app/shared/substitutor/label-substitutor.directive.spec.ts
@@ -423,7 +423,7 @@ describe('LabelSubstitutorDirective', () => {
             expect(labelSubstitutionService.substituteLabel).toHaveBeenCalledWith({ LabelB: '', LabelA: '7 Mar 2018' }, label);
         });
 
-        it('should pass form field value with invalid date', () => {
+        it('should pass form field value with invalid date when both form and case field values present', () => {
           let label = 'someLabel';
           comp.caseField = field('LabelB', '', {
               id: 'LabelB',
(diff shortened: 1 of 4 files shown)


src/app/shared/substitutor/label-substitutor.directive.spec.ts
```diff
@@ -420,8 +420,26 @@ describe('LabelSubstitutorDirective', () => {
             });
             fixture.detectChanges();
 
-            expect(labelSubstitutionService.substituteLabel).toHaveBeenCalledWith({ LabelB: '', LabelA: '07-03-2018' }, label);
+            expect(labelSubstitutionService.substituteLabel).toHaveBeenCalledWith({ LabelB: '', LabelA: '7 Mar 2018' }, label);
         });
+
+        it('should pass form field value with invalid date', () => {
```
andrewfolga: Should we be consistent with the test naming convention above?

andrewfolga:
Build is failing.

src/app/shared/substitutor/label-substitutor.directive.spec.ts
mario-paniccia: Done

PR #81, opened 2018-05-09
RDM-1851 - Header and Footer to be inline with the UI design

MKaramath:
https://tools.hmcts.net/jira/browse/RDM-1851

**_Dependency note_**:  This change has dependency with the change on ccd-case-ui-toolkit branch RDM-1851

The header should look like with similar dimensions as in: https://reform-pattern-library.herokuapp.com/
The existing menus should be moved to the black header part, as seen in the pattern library header with menu "Elements""components" etc.

Footer should look like with the similar design as shown in the UI design layout. 

Header and Footer will follow the GOV.IN standards in terms of the desktop screen width, etc.

```
[ ] Yes
[x] No
```


src/app/core/core.component.html
```diff
@@ -1,11 +1,18 @@
...
-<cut-nav-bar class="global-navigation">
-  <div class="nav-left" leftNavLinks>
-    <ul id="menu-links-left">
+                [isSolicitor]="isSolicitor()"
+                username="{{profile?.user.idam.forename | titlecase}} {{profile?.user.idam.surname | titlecase}}"
+                (signOutRequest)='signOut()'>
+
+  <!-- Header menu : Start -->
+  <!-- "true" condition is placeholder whether to display header menu or not -->
+  <div *ngIf="true" headerNavigation>
```
mario-paniccia: what if we introduced a method in the component displayHeader() that for the moment just returns true? It'll be more self explanatory and we could also remove the explanatory comment

PR #87, opened 2018-05-10
Rdm 2055 Case mgmt web: Consume and use new column

andrewfolga:
https://tools.hmcts.net/jira/browse/RDM-2055

### Change description ###
Case mgmt web: Consume and use new column

```
[ ] Yes
[x] No
```


infrastructure/main.tf
```diff
@@ -15,9 +15,14 @@ locals {
...
   // Vault name
-  previewVaultName = "${var.product}-case-web"
+  previewVaultName = "${var.product}-mgmt-web"
   nonPreviewVaultName = "ccd-case-web-${var.env}"
   vaultName = "${(var.env == "preview" || var.env == "spreview") ? local.previewVaultName : local.nonPreviewVaultName}"
+
+  // Vault URI
+  previewVaultUri = "https://ccd-case-web-aat.vault.azure.net/"
+  nonPreviewVaultUri = "${module.ccd-case-management-web-vault.key_vault_uri}"
+  vaultUri = "${(var.env == "preview" || var.env == "spreview") ? local.previewVaultUri : local.nonPreviewVaultUri}"
```
mario-paniccia: === slightly better?

src/app/core/order/order.service.spec.ts
```diff
@@ -16,24 +16,24 @@ describe('OrderService', () => {
   describe('compareAsc', () => {
 
     it('should sort in ascending order', () => {
-      expect(orderService.compareAsc(ITEM_1, ITEM_2)).toBe(-1);
```
mario-paniccia: how come this change? According to the describe this is specifically testing the compareAsc method. Also, sortAsc is deprecated

src/app/core/order/order.service.spec.ts
```diff
@@ -16,24 +16,24 @@ describe('OrderService', () => {
   describe('compareAsc', () => {
```
mario-paniccia: sortAsc?

src/app/core/order/order.service.ts
```diff
@@ -8,21 +8,9 @@ export class OrderService {
...
-   * @returns {Orderable[]} Sorted clone array.
-   */
-  sort<T extends Orderable>(array: T[]): T[] {
-    return array
-      .slice(0)
-      .sort(this.compareAsc);
-  }
-
-  compareAsc(a: Orderable, b: Orderable): number {
+  private static readonly DEFAULT_COMPARE_FUNCTION = function compareFunction(a: Orderable, b: Orderable): number {
```
mario-paniccia: is it possible to use the newer ES6 Arrow notation?

src/app/shared/case-editor/case-edit-submit.component.ts
```diff
@@ -32,14 +33,30 @@ export class CaseEditSubmitComponent implements OnInit {
   triggerText: string = CallbackErrorsComponent.TRIGGER_TEXT_SUBMIT;
   wizard: Wizard;
   profile: Profile;
+  sortedFields: CaseField[];
+
+  public static readonly SHOW_SUMMARY_CONTENT_COMPARE_FUNCTION = function compareFunction(a: CaseField, b: CaseField): number {
```
mario-paniccia: is it possible to use the newer ES6 Arrow notation?

src/app/shared/case-editor/case-edit-submit.component.ts
```diff
@@ -124,15 +142,28 @@ export class CaseEditSubmitComponent implements OnInit {
...
     }
 
-    // found no fields to show in summary page
+    // found no fields to show in CYA summary page
+    return false;
+  }
+
+  readOnlySummaryfieldsToDisplayExists(): boolean {
+
+    for (let eventCaseField of this.eventTrigger.case_fields) {
```
mario-paniccia: would this work?

return this.eventTrigger.case_fields.some(field => field.show_summary_change_option)

mario-paniccia: perhaps we could also move this method on the CaseEventTrigger itself. Not sure though we are able to do that at the moment with the way we deserialise JSON objects

src/app/shared/case-editor/case-edit-submit.html
```diff
@@ -55,6 +55,31 @@ <h2 class="plain-message">Check the information below carefully. This will form
       </tbody>
     </table>
   </ng-container>
+  <ng-container *ngIf="readOnlySummaryfieldsToDisplayExists()">
+    
+    <table class="form-table">
+      <tbody>
+        <ng-container *ngFor="let field of this.sortedFields">          
+            <tr *ngIf="!(field | ccdIsCompound); else ReadOnlyCompoundRow" ccdLabelSubstitutor [caseField]="field"
```
mario-paniccia: can we use NgSwitch instead of this construct? that's the recommended angular approach for this scenarios and I find it much more easy to understand. There's an example in case-viewer.component

src/app/shared/case-editor/case-edit-submit.component.ts
```diff
@@ -32,14 +33,30 @@ export class CaseEditSubmitComponent implements OnInit {
   triggerText: string = CallbackErrorsComponent.TRIGGER_TEXT_SUBMIT;
   wizard: Wizard;
   profile: Profile;
+  sortedFields: CaseField[];
```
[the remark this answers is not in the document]
andrewfolga: Shouldn't we avoid too long names? I can make the function that sets this field called more descriptive.

src/app/core/order/order.service.ts
```diff
@@ -36,4 +24,17 @@ export class OrderService {
...
+
+  /**
+   * Clone and sort array. Ascending order used by default.
+   *
+   * @param array Array to sort
+   * @returns {Orderable[]} Sorted clone array.
+   */
+  sort<T extends Orderable>(array: T[], sortingFunction = this.sortAsc): T[] {
+    return array
+      .slice(0)
```
mario-paniccia: can we remove 0 and leave .slice() ?
0 is the default

src/app/shared/case-editor/case-edit-submit.component.ts
andrewfolga: Just did the first comment for now.

src/app/shared/case-editor/case-edit-submit.component.ts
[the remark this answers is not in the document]
mario-paniccia: but the variable could still be mistakenly used in the future because it's name is misleading and cause bugs

PR #98, opened 2018-05-16
Rdm 1149 DateTime Support

tzarsmango:
https://tools.hmcts.net/jira/browse/RDM-1149
This one is for adding additional hour, minute and second boxes to existing Date component

This depends on
- https://github.com/hmcts/ccd-case-ui-toolkit/pull/12
- https://github.com/hmcts/ccd-definition-store-api/pull/28
- https://github.com/hmcts/ccd-data-store-api/pull/65

```
[ ] Yes
[X] No
```


src/app/shared/palette/date/write-date-field.component.ts
```diff
@@ -14,4 +14,8 @@ export class WriteDateFieldComponent extends AbstractFieldWriteComponent impleme
     this.dateControl = this.registerControl(new FormControl(this.caseField.value));
   }
 
+  isDateTime(): boolean {
+    return this.caseField.field_type.id === 'DateTime';
```
mario-paniccia: field_type has got an 'id' and a 'type' property. I've seen sometime we use 'id' but mainly we use 'type' to check the type
to me feels more natural .type

stubs/unit/aggregated.api.json
```diff
@@ -35,1666 +35,1671 @@
     }
   },
   "case-types": [
```
mario-paniccia: it could be possible to have an example of DateTime in the aggregated api? ignore me if it's there, I couldn't find it

src/app/shared/palette/date/write-date-field.html
```diff
@@ -8,6 +8,7 @@
     </legend>
 
     <cut-date-input [id]="id()"
+                    [dateTime]="isDateTime()"
```
mario-paniccia: wonder whether this property is really needed. Can't the cut-date-input infer it's a date time by looking at the presence of the time part in the raw string?

src/app/shared/palette/date/write-date-field.component.ts
tzarsmango: id and type differ for some types like AddressUK, AddressGlobal & AddressGlobalUK. For these types id is AddressUK while the type is Complex. For this one both mean the same...

PR #107, opened 2018-05-21
Rdm 2208 - Previously entered data is retained and the 'Continue' button is disabled

tzarsmango:
https://tools.hmcts.net/jira/browse/RDM-2208

This one is for fixing mandatory fields preventing to continue when a child field of a complex field is hidden.

```
[ ] Yes
[X] No
```

### DIFF
--- a/src/app/shared/conditional-show/conditional-show.directive.ts
+++ b/src/app/shared/conditional-show/conditional-show.directive.ts
@@ -130,6 +130,7 @@ export class ConditionalShowDirective implements AfterViewInit, OnDestroy {
     }
   }
 
+  // TODO This must be extracted to a generic service for traversing see RDM-2233
   private checkHideShowCondition(key: string, aControl: AbstractControl) {
     if (aControl instanceof FormArray) {  // We're in a collection
       // console.log('traversing array', aControl);
--- a/src/app/shared/conditional-show/conditional-show.directive.ts
+++ b/src/app/shared/conditional-show/conditional-show.directive.ts
@@ -142,8 +142,8 @@ export class ConditionalShowDirective implements AfterViewInit, OnDestroy {
       if (aControl.get('value')) { // Complex Field
         let complexControl = aControl.get('value') as FormGroup;
         Object.keys(complexControl.controls).forEach(controlKey => {
-          // console.log('traversing formGroup item', key, aControl.get('value').get(key));
-          this.checkHideShowCondition(controlKey, aControl.get('value').get(controlKey));
+          // console.log('traversing formGroup item', key, complexControl.get(key));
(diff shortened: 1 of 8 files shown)


src/app/shared/conditional-show/conditional-show.directive.ts
```diff
@@ -116,4 +129,33 @@ export class ConditionalShowDirective implements AfterViewInit, OnDestroy {
...
+        // console.log('in array', formControl);
+        this.checkHideShowCondition('' + i, formControl)
+      });
+    } else if (aControl instanceof FormGroup) {
+      // console.log('met a FormGroup ', aControl, ' fromGroup.controls', aControl.controls);
+      if (aControl.get('value')) { // Complex Field
+        let complexControl = aControl.get('value') as FormGroup;
+        Object.keys(complexControl.controls).forEach(controlKey => {
+          // console.log('traversing formGroup item', key, aControl.get('value').get(key));
+          this.checkHideShowCondition(controlKey, aControl.get('value').get(controlKey));
```
andrewfolga: Can we replace aControl with complexControl here?

andrewfolga:
Also going one page forward and coming back causes the registrar to readd the same show condition directives again and again.

src/app/shared/conditional-show/conditional-show.directive.ts
```diff
@@ -116,4 +129,33 @@ export class ConditionalShowDirective implements AfterViewInit, OnDestroy {
       this.formChangesSubscription.unsubscribe();
     }
   }
+
+  private checkHideShowCondition(key: string, aControl: AbstractControl) {
```
andrewfolga: Just a thought but was wondering how difficult would it be to make this a separate formControlTraverser passing in a condition of the control (in this case `if (aControl.invalid)`) and action to take once condition is met (in this case `this.registry.refresh()`).

tzarsmango: Created https://tools.hmcts.net/jira/browse/RDM-2233 for this purpose...

src/app/shared/conditional-show/conditional-show.directive.ts
```diff
@@ -1,13 +1,14 @@
 import { AfterViewInit, Directive, ElementRef, Input, OnDestroy } from '@angular/core';
```
mario-paniccia: @tzarsmango haven't checked you PR yet. But I just noticed that we use the following method in this file:

 private getFormFieldsValuesIncludingDisabled() {
    return this.formGroup.getRawValue();
  }

I don't remember exactly the reason I decided to include also disabled fields when initially developed show condition, but, have you tried changing this method to use .value property rather than .getRawValue() in order to exclude disabled fields? Just worth mentioning, it might not work

PR #114, opened 2018-05-23
RDM-1914: Redesigning collection button placement

MKaramath:
https://tools.hmcts.net/jira/browse/RDM-1914

### Change description ###
This change is to redesign the collection button placements as follows:

**Add button changes**
1. The Add button label is changed from 'Add item' to 'Add new'. 
2. At the beginning, only 1 Add button will appear, but when 1 or more items are added to the collection then a new Add button will appear just below the collection header.

**Remove button changes**
1. The Remove button label is changed from 'Remove item' to 'Remove'.
2. The Remove button is moved to appear on the right of the collection title.
3. When Remove button is clicked, the dialog will appear to confirm the removal of item with button 'Remove' and 'Cancel'.
4. If 'Remove' is clicked on the dialog, then the item will be deleted or else if 'Cancel' is clicked then the item will not be deleted.

```
[ ] Yes
[x] No
```


package.json
```diff
@@ -30,6 +30,8 @@
   },
   "dependencies": {
     "@angular/animations": "~5.0.0",
+    "@angular/cdk": "^5.0.0",
+    "@angular/material": "^5.0.0",
```
mario-paniccia: Have we checked with @vlaurin to make sure it's ok to introduce Angular material?

MKaramath: Yes, checked with @vlaurin and he is happy to use Angular material.

src/app/shared/palette/collection/write-collection-field.component.ts
```diff
@@ -72,4 +74,27 @@ export class WriteCollectionFieldComponent extends AbstractFieldWriteComponent i
     return index ? `${this.caseField.label} ${displayIndex}` : this.caseField.label;
   }
 
+  openModal(i: number) {
```
mario-paniccia: can we give i a more meaningful name?

src/app/shared/palette/collection/write-collection-field.component.ts
```diff
@@ -72,4 +74,27 @@ export class WriteCollectionFieldComponent extends AbstractFieldWriteComponent i
...
+    const dialogConfig = new MatDialogConfig();
+    dialogConfig.disableClose = true;
+    dialogConfig.autoFocus = true;
+    dialogConfig.ariaLabel = 'Label';
+    dialogConfig.height = '220px';
```
mario-paniccia: can this be moved to a css file?

MKaramath: I tried, but since this is just a configurable object which takes these values.  So, unfortunately it is not possible put them in css file.

src/app/shared/palette/collection/write-collection-field.component.ts
MKaramath: i is basically the index of the items within the collection type field.

PR #118, opened 2018-05-24
RDM-1457 Show only Address Entry Fields while searching

tzarsmango:
https://tools.hmcts.net/jira/browse/RDM-1457

This one is for seeing only the Address fields and not the postcode lookup on search screen and caselist filter criteria 
```
[ ] Yes
[X] No
```

### DIFF
--- a/src/app/shared/palette/address/write-address-field.html
+++ b/src/app/shared/palette/address/write-address-field.html
@@ -23,10 +23,10 @@
     </select>
   </div>
 
-  <span class="manual-link" *ngIf="!isAddressSet()" (click)="blankAddress()">I can't enter a UK postcode</span>
+  <span class="manual-link" *ngIf="!shouldShowDetailFields()" (click)="blankAddress()">I can't enter a UK postcode</span>
 
   <ccd-write-complex-type-field
-    [hidden]="!isAddressSet()"
+    [hidden]="!shouldShowDetailFields()"
     [(caseField)]="caseField"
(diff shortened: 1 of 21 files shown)


src/app/shared/palette/address/write-address-field.html
```diff
@@ -2,7 +2,7 @@
 
   <h3 class="heading-medium">{{caseField | ccdFieldLabel }}</h3>
 
-  <div class="form-group" id="postcodeLookup" [ngClass]="{'form-group-error': missingPostcode}">
+  <div class="form-group" id="postcodeLookup" [ngClass]="{'form-group-error': missingPostcode}" *ngIf="!isSearchField">
```
mario-paniccia: can we rename this property to showPostcode? this way it's more generic. If in the future we have other sections of the ui where we don't want to show the postcode we wouldn't have to change the code again

src/app/shared/palette/base-field/field-write.component.ts
```diff
@@ -54,6 +55,12 @@ export class FieldWriteComponent extends AbstractFieldWriteComponent implements
     if (this.caseField.field_type.id === 'AddressGlobal') {
       component.instance['ignoreMandatory'] = true;
     }
+    if (this.isSearchField && (
```
mario-paniccia: mmm I have some doubts on this. This breaks encapsulation, field-write.component.ts should be generic and know nothing about the existence of a Address type

PR #122, opened 2018-05-31
case-activity-url-parameter-switch

agilks:
This is a change to update configuration in order to toggle on and off the use of the case activity service by the front end. The change mean it is only enabled for a specific test environment.

```
[ ] Yes
[x] No
```


agilks:
Not required

PR #117, opened 2018-05-24
RDM-1873: H1, H2, Label text throughout configuration

MKaramath:
https://tools.hmcts.net/jira/browse/RDM-1873
**_Note_**: This PR should be merge along with the below PR https://github.com/hmcts/ccd-case-management-web/pull/137

### Change description ###

The page label, event trigger name and case reference id configuration are changed.  

**Line1**
H1(24px): <Page label>(optional in caseEventToField tab)<space>"-"<space><Event name>(mandatory defined in CaseEvent tab)
**Line2** 
H2(19px): Key element which is Case reference number currently

**Example**
Petitioner details - Adding party details (line 1)
#1305-5234-2345-1234 (line 2)
First name:

**Does this PR introduce a breaking change?** (check one with "x")

```
[ ] Yes
[x] No
```

PR #137, opened 2018-06-08
RDM-1662: Create Dropdown and other UI fixes

MKaramath:
https://tools.hmcts.net/jira/browse/RDM-1662

**_Note_**: This PR should be merge along with the below PR https://github.com/hmcts/ccd-case-management-web/pull/117

### Change description ###
The dropdown fields to conform with UX guidelines and behaviour.   This change is applied on all pages where dropdown field exists. 

Other UI changes:
- Responsive is applied on Case list, Create case pages
- Dropdown styling is corrected to guidelines on all browsers
- On case list, the spaces surrounding the case list filters is fixed with responsiveness; the left and right alignments of filter results are corrected.
- On create case, headings (H1, H2, H3), field labels, label hints, and all others fields styling are corrected to conform the UX guidelines
- On search filter, the dropdowns are updated as per guidelines
- **£** symbol of money input field is now placed fixed inside the text as per guidelines
- Read only email field on the wizards pages is enabled to a link
- The changes related to the PR https://github.com/hmcts/ccd-case-management-web/pull/117 is now merged into this PR due to the UI layout dependency.
- The read only email field is enabled as link to open a blank email addressing to the email (refer https://tools.hmcts.net/jira/browse/RDM-2309)

```
[ ] Yes
[x] No
```


src/app/shared/case-editor/case-edit-submit.component.ts
```diff
@@ -223,4 +223,20 @@ export class CaseEditSubmitComponent implements OnInit {
       .filter(cf => cf.show_summary_content_option);
   }
 
+  getCaseId(): String {
+    return (this.caseEdit.caseDetails ? this.caseEdit.caseDetails.case_id : '');
+  }
+
+  isEmail(field: CaseField): boolean {
```
mario-paniccia: not sure I got what this method is doing. Why we need to call this.summaryCaseField(field)?

src/app/shared/case-editor/case-edit-submit.html
```diff
@@ -28,13 +33,18 @@ <h2 class="plain-message">Check the information below carefully. This will form
             <ng-container *ngIf="canShowFieldInCYA(field)">
               <tr ccdConditionalShow ccdLabelSubstitutor [caseField]="field"
                 [formGroup]="editForm.controls['data']" [eventFields]="eventTrigger.case_fields">
-                <th class="valign-top">{{field.label}}</th>
+                <th class="valign-top"><span class="text-16">{{field.label}}</span></th>
                 <td class="form-cell">
-                  <ccd-field-read [caseField]="summaryCaseField(field)" [context]="paletteContext"></ccd-field-read>
+                  <a href="mailto:{{summaryCaseField(field).value}}" class="text-16" *ngIf="isEmail(field)">
```
mario-paniccia: this should be changed. It's not responsibility of case-edit-submit.html to check whether a field has an email and display that accordingly. 
This should be moved inside the email read component template

src/app/shared/case-editor/case-edit-submit.component.ts
MKaramath: I have fixed this by assigning it to a variable now.   I am calling this.summaryCaseField(field) to filter non-null fields.

src/app/shared/case-editor/case-edit-submit.html
MKaramath: I tried but it was bit challenging due to that all types of read only fields are rendering dynamically.  May be I will look at that to improve sometimes in future if I find a better solution.

mario-paniccia: becasuse theis field is used only on the check your answers section of this page, we should rename it to ccd-cya-field-read 
and we should move all this logic inside it's template. We should also move the isEmail method inside the ccdCYAFieldRead component
Karamath agree that he's going to open a new ticket to do this change

ngul:
This PR is absolutely massive. could there anything be done to split it?

PR #141, opened 2018-06-13
RDM-597 - Case History

gbenadikar:
https://tools.hmcts.net/jira/browse/RDM-597

### Change description ###
Introduced new component for showing case history for specific event. Designs not available yet, simplistic view for now to show all case details in single page. These changes have a dependency on ccd-data-store-api PR https://github.com/hmcts/ccd-data-store-api/pull/95

```
[ ] Yes
[x] No
```


src/app/app.routing.ts
```diff
@@ -131,6 +133,13 @@ const routes: Routes = [
                 component: CaseEditPageComponent,
               }
             ]
+          },
+          {
+            path: 'event/:eid',
```
mario-paniccia: I find this path a bit confusing because we are not entering an event, but the case history associated to that event. 
Shall we change this to something like:
`event/:eid/history`

src/app/core/cases/case-history-view.model.ts
```diff
@@ -0,0 +1,18 @@
+import { CaseTab } from './case-tab.model';
+import { CaseViewEvent } from './case-view-event.model';
+
+export class CaseHistoryView {
```
mario-paniccia: can we just call it CaseHistory?

src/app/core/cases/cases.service.spec.ts
```diff
@@ -425,4 +429,68 @@ describe('CasesService', () => {
         });
     });
   });
+
+  describe('getCaseHistoryView()', () => {
+    const EVENT: CaseViewEvent = {
```
mario-paniccia: recently we moved away from direct instantiation of objects in tests with the object literal notation because makes the test very fragile and reduces maintainability. 
We try to define fixture files with creational methods that can be reused by many tests. Can you have a look at:
`profile.test.fixture.ts`

src/app/core/cases/cases.service.ts
```diff
@@ -168,6 +169,26 @@ export class CasesService {
       });
   }
 
+  getCaseHistoryView(jurisdictionId: string,
```
mario-paniccia: CasesService is in my opinion growing too much with too many responsibility. 
Can we introduce a CaseHistoryService and move this method there?

src/app/core/cases/cases.service.ts
```diff
@@ -168,6 +169,26 @@ export class CasesService {
...
+              caseTypeId: string,
+              caseId: string,
+              eventId: string): Observable<CaseHistoryView> {
+    const url = this.appConfig.getApiUrl()
```
mario-paniccia: can we move the construction of the url inside a appConfig.getCaseHistoryUrl()?

src/app/shared/case-history/case-history.resolver.ts
```diff
@@ -0,0 +1,45 @@
...
+import { Observable } from 'rxjs/Observable';
+import { CasesService } from '../../core/cases/cases.service';
+import { Response } from '@angular/http';
+import 'rxjs/add/operator/catch';
+import { CaseHistoryView } from '../../core/cases/case-history-view.model';
+
+@Injectable()
+export class CaseHistoryResolver implements Resolve<CaseHistoryView> {
+
+  public static readonly PARAM_JURISDICTION_ID = 'jid';
```
mario-paniccia: private?

src/app/core/cases/cases.service.ts
gbenadikar: Done

src/app/core/cases/cases.service.ts
gbenadikar: Done

src/app/core/cases/cases.service.spec.ts
gbenadikar: Done for case history

src/app/core/cases/case-history-view.model.ts
gbenadikar: done

src/app/app.routing.ts
gbenadikar: yeah, history looks more logical, done.

src/app/shared/case-history/case-history.resolver.ts
gbenadikar: requires access in the test file, hence public.

gbenadikar:
Completed code review changes asked.

src/app/core/cases/case-history.model.ts
```diff
@@ -0,0 +1,23 @@
...
+    name: string,
+    description?: string,
+    jurisdiction: {
+      id: string,
+      name: string,
+      description?: string
+    }
+  };
+
+  @Type(() => CaseTab)
```
mario-paniccia: 👍

src/app/core/cases/case-history.model.ts
```diff
@@ -0,0 +1,23 @@
...
+  tabs: CaseTab[];
+
+  @Type(() => CaseViewEvent)
```
mario-paniccia: 👍

src/app/core/cases/case-history.service.ts
```diff
@@ -0,0 +1,32 @@
...
+    const url = this.appConfig.getCaseHistoryUrl(jurisdictionId, caseTypeId, caseId, eventId);
+
+    return this.httpService
+      .get(url)
+      .map(response => response.json())
+      .catch((error: any): any => {
+        this.httpErrorService.setError(error);
+        return Observable.throw(error);
+      })
+      .map((caseHistory: Object) => plainToClass(CaseHistory, caseHistory));
```
mario-paniccia: way to go 👍

src/app/core/cases/case-history.model.ts
```diff
@@ -0,0 +1,23 @@
...
+import { CaseViewEvent } from './case-view-event.model';
+import { Type } from 'class-transformer';
+
+export class CaseHistory {
+  case_id?: string;
+  case_type: {
+    id: string,
+    name: string,
+    description?: string,
+    jurisdiction: {
```
mario-paniccia: can't we reuse our Jurisdiction type here?
```jurisdiction: Jurisdiction```
we could also see if automatic deserialising works:
```
@Type(() => Jurisdicition)
jurisdiction: Jurisdiction
```

gbenadikar: It didn't allow Type declaration inside the case type literal declaration so moved case history case type as a separate class and reuse Jurisdiction

PR #146, opened 2018-06-18
RDM-2325: Update PREVIEW infrastructure variables

MKaramath:
https://tools.hmcts.net/jira/browse/RDM-2325

### Change description ###

Update the preview infrastructure variables to enable the authentications

```
[ ] Yes
[x] No
```


infrastructure/preview.tfvars
```diff
@@ -1,3 +1,7 @@
+idam_authentication_web_url = "https://idam.preprod.ccidam.reform.hmcts.net"
```
vlaurin: Why do we have `idam_authentication_web_url` and `authentication_web_url`? I believe one of them is not used and should be removed

MKaramath: authentication_web_url is now removed.

PR #147, opened 2018-06-19
RDM-2307: Are you sure when clicking browser button

MKaramath:
https://tools.hmcts.net/jira/browse/RDM-2307

### Change description ###
When a document has already been attached to a Document Field - some users are clicking browser again to update a new document.

Users would like a warning saying

"Are you user you wish to replace the existing document ? " yes / no.

If yes then carry on and allow a file to be selected
If No cancel the operation and leave the original file as is.

IF a file has never been attached to the field then the message should not display

```
[ ] Yes
[x] No
```


MKaramath:
The new PR is created for this change and merged in the master.   So closing this PR as duplicate. 
https://github.com/hmcts/ccd-case-management-web/pull/160

PR #164, opened 2018-07-04
RDM-2416 Post Order Summary to Service on submit callback

andrewfolga:
https://tools.hmcts.net/jira/browse/RDM-2416

### Change description ###
Order Summary Post to Service on submit callback

```
[ ] Yes
[x] No
```


src/app/shared/palette/order-summary/write-order-summary-field.html
```diff
@@ -0,0 +1,3 @@
+<div>
```
vlaurin: Do we need the empty `div` around the component?

src/app/shared/palette/order-summary/write-order-summary-field.component.ts
```diff
@@ -0,0 +1,40 @@
...
+})
+export class WriteOrderSummaryFieldComponent extends AbstractFieldWriteComponent implements OnInit {
+
+  paymentReference: FormControl;
+  paymentTotal: FormControl;
+  feesArray: FormArray;
+  orderSummaryGroup: FormGroup;
+
+  ngOnInit(): void {
+    this.orderSummaryGroup = this.registerControl(new FormGroup({}));
```
vlaurin: Could a comment be added just to explain quickly why those are implemented manually rather than making use of `WriteComplexFieldComponent`? Basically, that given the view is read-only, the tree of form controls is not being built automatically and need to be built manually instead.

src/app/shared/palette/order-summary/write-order-summary-field.component.ts
```diff
@@ -0,0 +1,40 @@
...
+@Component({
+  selector: 'ccd-write-order-summary-field',
+  templateUrl: './write-order-summary-field.html'
+})
+export class WriteOrderSummaryFieldComponent extends AbstractFieldWriteComponent implements OnInit {
+
+  paymentReference: FormControl;
+  paymentTotal: FormControl;
+  feesArray: FormArray;
+  orderSummaryGroup: FormGroup;
```
vlaurin: I don't believe exposing those variables at class level is needed, they could be kept in function-scope.

PR #165, opened 2018-07-05
RDM-1240 - search by metadata

gbenadikar:
https://tools.hmcts.net/jira/browse/RDM-1240

### Change description ###
Search filter component and search component changed to support search on metadata fields. Have retained some code to support search based on state for the case list component.
Depends on
https://github.com/hmcts/ccd-definition-store-api/pull/79
https://github.com/hmcts/ccd-data-store-api/pull/116

```
[ ] Yes
[x] No
```


src/app/shared/search/filters/search-filters.component.ts
```diff
@@ -59,9 +60,18 @@ export class SearchFiltersComponent implements OnInit {
...
     this.selected.formGroup = this.formGroup;
     this.selected.page = 1;
+    this.selected.metadataFields = this.getMetadataFields();
     this.onApply.emit(this.selected);
   }
 
+  getMetadataFields(): string[] {
+    if (this.searchInputs) {
+      return this.searchInputs.filter(searchInput => searchInput.field.metadata === true).map(function (searchInput) {
+        return searchInput.field.id;
```
vlaurin: Seems a bit weird to switch from lambda to function halfway through. Also `.map` should be more legible if aligned below `.filter`.

```ts
return this.searchInputs.filter(searchInput => searchInput.field.metadata === true)
                        .map(searchInput => searchInput.field.id);
````

gbenadikar: Right :). Updated now

PR #167, opened 2018-07-12
RDM-2323 Allow value to be substitutable from Search Result View

adr1ancho:

RDM-2323

### Change description ###

Allow token to be substitutable from Search Result View

```
[ ] Yes
[x] No
```


src/app/shared/search/search-result.component.ts
```diff
@@ -89,12 +91,44 @@ export class SearchResultComponent implements OnChanges {
...
 
+  /**
+   * Hydrates result view with case field definitions.
+   */
+  // A longer term resolution is to move this piece of logic to the backend
+  hydrateResultView(): void {
+    this.resultView.results.forEach(result => {
+      const caseFields = [];
+
+      Object.keys(result.case_fields).forEach(fieldId => {
```
mario-paniccia: can we introduce a method 
`getCaseFieldsId()` on SearchResultViewItem and use it here: 
```
result.getCaseFieldsId().forEach(fieldId => ....

Comment (mario-paniccia) [inline src/app/shared/search/search-result.component.ts]: I think we could iterate this way for a simpler solution:

```
 for (let [fieldId, field] of Object.entries(result.case_fields)) {
...
      }
```

File: src/app/shared/search/import { CaseField } from '..:domain:definition:ca [hunk 2]
```diff
@@ -0,0 +1,8 @@
+import { CaseField } from '../domain/definition/case-field.model';
```
mario-paniccia: don't understand what this file is. seems a copy of the one below: src/app/shared/search/search-result-view-item.model.ts

src/app/shared/palette/label/label-field.html
```diff
@@ -1 +1 @@
-<dl class="case-field"><dt><ccd-markdown [content]="caseField.label"></ccd-markdown></dt><dd></dd></dl>
+<dl class="case-field"><dt><ccd-markdown [content]="caseField.value || caseField.label"></ccd-markdown></dt><dd></dd></dl>
```
mario-paniccia: aren't we breaking encapsulation of the caseField here? Ideally the caseField itself should know how to return it's label and where to get it from. 
Can we expose a method
`
getLabel() {
 return caseField.value || caseField.label
}
`
?

mario-paniccia: aren't we breaking encapsulation here? it's the caseField itself that should know how to return it's label. Can we add a method in caseField:
`getLabel() {
  return caseField.value || caseField.label
}



label-substitutor.directive.ts

src/app/shared/search/import { CaseField } from '..:domain:definition:ca
```diff
@@ -0,0 +1,8 @@
...
+
+export class SearchResultViewItem {
+  case_id: string;
+  case_fields: object;
+  hydrated_case_fields?: CaseField[];
+  columns?: object;
```
mario-paniccia: can we have `columns` declared with a type? We could use ES6 Map

src/app/shared/search/search-result.component.ts
```diff
@@ -89,12 +91,44 @@ export class SearchResultComponent implements OnChanges {
...
     if (changes['page']) {
       this.selected.page = (changes['page']).currentValue;
     }
   }
 
+  /**
+   * Hydrates result view with case field definitions.
+   */
+  // A longer term resolution is to move this piece of logic to the backend
+  hydrateResultView(): void {
```
mario-paniccia: it took me a while to figure out what's going on here. I think this new code could use some refactoring and better naming to make it more self explanatory. 
Also, I think instead of adding new fields to SearchResultViewItem (hydrated_case_fields and columns), we could consider introducing a new concept/type in our domain, for example DisplayedSearchItem. And we 
could have SearchResultComponent create list of DisplayedSearchItem from the SearchResultViewItem.
The SearchResultComponent view could then operate on the DisplayedSearchItem only.
This way we keep SearchResultViewItem being a one to one mapping with the data returned from the backend api.

src/app/shared/search/search-result.component.ts
```diff
@@ -89,12 +91,44 @@ export class SearchResultComponent implements OnChanges {
...
+        caseFields.push({
+          id: fieldId,
+          label: null,
+          field_type: {},
+          value: field,
+          display_context: null,
+        });
+      });
+
+      result.hydrated_case_fields = caseFields;
```
mario-paniccia: can we find a better naming than hydrated_case_fields? it's difficult to understand without looking at the code what's the purpose of this array of CaseFields

src/app/shared/palette/label/label-field.html
vlaurin: Nope, that knowledge doesn't belong to the `CaseField`, it's specific to the `LabelField`. We certainly don't want other types of fields to start returning their values as a label.

vlaurin: Same as above, this is an exception of the `LabelField`; this is not something that should be done for any other type of fields. As such, this logic cannot belong to `CaseField`.

src/app/shared/search/import { CaseField } from '..:domain:definition:ca
vlaurin: 👍

src/app/shared/search/search-result.component.ts
vlaurin: 👍

src/app/shared/search/search-result.html
```diff
@@ -21,13 +21,8 @@
...
     </td>
     <td *ngFor="let col of resultView.columns">
-      <div class="text-16" [style.visibility]="hideRows ? 'hidden' : 'visible'">
-        <ccd-field-read [caseField]="{
-                id: col.case_field_id,
-                label: col.label,
-                field_type: col.case_field_type,
-                value: result.case_fields[col.case_field_id]
-          }"></ccd-field-read>
+      <div class="text-16" [style.visibility]="hideRows ? 'hidden' : 'visible'" ccdLabelSubstitutor [caseField]="result.columns[col.case_field_id]" [eventFields]="result.hydrated_case_fields">
```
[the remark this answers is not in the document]
vlaurin: 👍

src/app/shared/search/search-result.component.ts
mario-paniccia: what's the reason for not having here: 
```result.hydrated_case_fields = result.case_fields```
I'm sure there is a reason but can't get it :)

src/app/shared/palette/label/label-field.html
mario-paniccia: That's because this particular way of retrieving the label is specific only for the search view right?

mario-paniccia: in theory then we could have a method getSearchResultLabel() on the CaseField

mario-paniccia: even though I don't like it much either :)
But doing `[content]="caseField.value || caseField.label"` smells a bit to me :)

vlaurin: I don't think leaking that logic anywhere outside of `LabelField`, which is the only one responsible for it, is good.
At best, I'd say we can introduce a `getLabel()` function inside `LabelField` to extract that from the template.

src/app/shared/search/search-result.component.ts
vlaurin: `hydrated_case_fields` is an array, while `result.case_fields` is an object used as a map. 
Because `ccdLabelSubstitutor` is expecting an array of fields, not a map, we're converting it to an array.

vlaurin: and `result.case_fields` only contains the raw values, not CaseField instances like `ccdLabelSubstitutor` is expecting

src/app/shared/search/search-result.component.ts
vlaurin: Definitely moving that logic to a service is a must. Everything you mentioned @mario-paniccia is good I think; though for me the proper fix would be to move the whole label substitution for read-only views to the backend. This would remove the need for most of the code introduced by this PR and try and prevent the complexity of the FE from increasing.

src/app/shared/search/search-result.component.ts
mario-paniccia: I see. What about a method 
getCaseFieldsArray(): CaseField[] on SearchResultViewItem itself then?

src/app/shared/search/search-result.component.ts
mario-paniccia: @vlaurin that's true yes. And I think we should do both. Because my suggestion is not strictly related to label substitution. But it's more an improvement to the design of the search component in general

adr1ancho:
Stale PR

PR #169, opened 2018-07-13
[PREVIEW] RDM-2386: Users using non supported browsers should not be redirected…

MKaramath:
https://tools.hmcts.net/jira/browse/RDM-2386

### Change description ###
If a user is using a non supported browser, he / she should be navigated to a page with the link :
https://www.gov.uk/help/browsers.

Our supported browser list:
https://tools.hmcts.net/confluence/display/RCCD/CCD+-+Supported+Browsers

```
[ ] Yes
[x] No
```


src/app/core/ccd-browser-support/ccd-browser-support.component.html
```diff
@@ -0,0 +1,13 @@
+<div class="browser-support" [class.screen-990]="isSolicitor" [class.hide]="!showUnsupportedBrowser">
```
mario-paniccia: can we use a ngIf here instead of [class.hide] ? seems a bit simpler to me with ngIf. We don't need to define an extra class

src/app/core/core.component.ts
```diff
@@ -52,4 +54,37 @@ export class CoreComponent implements OnInit, OnDestroy {
   isSolicitor(): boolean {
     return this.profile.isSolicitor();
   }
+
+  isUnsupportedBrowser(): boolean {
```
mario-paniccia: this method should not be on the core component. It would be better if we could move this into the ccd-browser-support.component itself

src/app/core/core.component.ts
```diff
@@ -52,4 +54,37 @@ export class CoreComponent implements OnInit, OnDestroy {
...
+
+    let ua = window.navigator.userAgent;
```
mario-paniccia: using `window` global object directly in Angular is not recommended. Any way we could avoid this?

mario-paniccia: for example have you considered using an angular library for device info detection: e.g.
https://github.com/KoderLabs/ngx-device-detector

src/app/core/core.component.ts
MKaramath: The scope will not satisfy if this method is moved into ccd-browser-support.component

src/app/core/core.component.ts
MKaramath: Done the change.  Thank you.

src/app/core/ccd-browser-support/ccd-browser-support.component.html
MKaramath: I have added class here to include additional styling in future such as animated transitions while hiding the band.

server.ts
```diff
@@ -29,11 +29,16 @@ const CONFIG = {
...
   'remote_print_service_url': process.env['PRINT_SERVICE_URL_REMOTE'] || 'https://return-case-doc.dev.ccd.reform.hmcts.net',
   'smart_survey_url': process.env['SMART_SURVEY_URL'] || 'https://www.smartsurvey.co.uk/s/CCDfeedback/',
+  'unsupported_browser_url': process.env['UNSUPPORTED_BROWSER_URL'] || 'https://www.gov.uk/help/browsers',
   'activity_url': process.env['CCD_ACTIVITY_URL'] || '',
   'activity_next_poll_request_ms': process.env['CCD_ACTIVITY_NEXT_POLL_REQUEST_MS'] || 5000,
   'activity_retry': process.env['CCD_ACTIVITY_RETRY'] || 5,
   'activity_batch_collection_delay_ms': process.env['CCD_ACTIVITY_BATCH_COLLECTION_DELAY_MS'] || 1,
-  'activity_max_request_per_batch': process.env['CCD_ACTIVITY_MAX_REQUEST_PER_BATCH'] || 25
+  'activity_max_request_per_batch': process.env['CCD_ACTIVITY_MAX_REQUEST_PER_BATCH'] || 25,
+  'chrome_version': process.env['CHROME_VERSION'] || 67,
```
mario-paniccia: can we rename these props to the format
min_required_<browser>_version 
?

src/app/core/core.component.ts
```diff
@@ -52,4 +56,24 @@ export class CoreComponent implements OnInit, OnDestroy {
...
+
+  isUnsupportedBrowser(): boolean {
+
+    let browser = this.deviceService.browser;
+    let browser_full_version = this.deviceService.browser_version;
+    let browser_version = parseInt(browser_full_version.substring(0, browser_full_version.indexOf('.')), 10);
+
+    switch (browser) {
+      case 'chrome':
+        return ( browser_version < this.appConfig.getChromeVersion() ) ? true : false;
```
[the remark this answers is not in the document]
MKaramath: Done

server.ts
MKaramath: Done.

Saravanabh:
This PR has been tested in all supported browser's. For unsupported browser, it was tested by chenaging the version in the local config file.
screen shot 2018-07-19 at 11 53 48

PR #170, opened 2018-07-16
RDM-1508 - allow metadata to be first column in case list/search results

gbenadikar:
https://tools.hmcts.net/jira/browse/RDM-1508

### Change description : ###
UI changes to show metadata field as first column for search results when configured in definition import as first column otherwise show case reference as first column. The import part of defining metadata fields in work basket result fields or search result fields has been already covered by previous PRs on story (RDM-1240)

```
[ ] Yes
[X] No
```


gbenadikar:
The changes are not required as there is another story that says flexible fields can be in the first column. I will raise a different PR for other story that will override this one. So no point in having this. Closing.

PR #157, opened 2018-06-29
Rdm 2223

tzarsmango:
https://tools.hmcts.net/jira/browse/RDM-2223

This PR is for saving the draft automatically when the user moves forward and backward in the create event wizard pages.

Should be merged after https://github.com/hmcts/ccd-data-store-api/pull/111

**Does this PR introduce a breaking change?** (check one with "x")

```
[ ] Yes
[X] No
```

PR #171, opened 2018-07-16
Rdm 2394

tzarsmango:
https://tools.hmcts.net/jira/browse/RDM-2394

Updated case list/ search result and case view pages to enable listing/viewing DRAFTS.

This PR contains the changes in this PR https://github.com/hmcts/ccd-case-management-web/pull/157

So must be merged after that PR

```
[ ] Yes
[X] No
```


src/app/cases/viewer/case-viewer.component.ts
```diff
@@ -105,4 +105,8 @@ export class CaseViewerComponent implements OnInit, OnDestroy {
       return acc.concat(tab.fields);
     }, []);
   }
+
+  isDraft(): boolean {
+    return this.caseDetails.case_id.length < 16;
```
mario-paniccia: in the resolver we use a different strategy to check whether a case is a draft, i.e. cid.startsWith(CaseResolver.DRAFT)
Can we use one approach only? 
Also, I would suggest introducing a method 
`isDraft(cid)`
somewhere, e.g. on CaseDetails,
 to avoid having this logic duplicated in multiple points

src/app/core/draft/draft.service.ts
```diff
@@ -0,0 +1,58 @@
...
+export class DraftService {
+  constructor(
+    private http: HttpService,
+    private appConfig: AppConfig,
+    private errorService: HttpErrorService
+  ) {}
+
+  createDraft(jid: string, ctid: string, eventData: CaseEventData): Observable<Draft> {
+    const saveDraftEndpoint = this.appConfig.getCaseDataUrl()
+      + `/caseworkers/:uid/jurisdictions/${jid}/case-types/${ctid}/event-trigger/${eventData.event.id}/drafts/`;
```
mario-paniccia: can we move this url construction logic inside a 
`appConfig.getDraftsUrl(jid, ctid, eventId)`
method?
This way you can reused it below also remove the duplication

src/app/cases/creator/case-creator-submit.component.ts
```diff
@@ -44,6 +47,29 @@ export class CaseCreatorSubmitComponent implements OnInit {
     return (sanitizedEditForm: CaseEventData) => this.casesService.validateCase(this.jurisdictionId, this.caseTypeId, sanitizedEditForm);
   }
 
+  private canSaveDraft(): boolean {
```
mario-paniccia: is it possible to move this method on 
`CaseEventTrigger`

mario-paniccia: you might have problem in making that work because of the way we deserialise objects, but let's give it a try?

src/app/cases/creator/case-creator-submit.component.ts
```diff
@@ -44,6 +47,29 @@ export class CaseCreatorSubmitComponent implements OnInit {
...
+    return this.eventTrigger.can_save_draft === true;
+  }
+
+  private isCreatingDraft(): boolean {
```
mario-paniccia: same as above

src/app/cases/creator/case-creator-submit.component.ts
mario-paniccia: also, how about we call this property 
`eventAllowsDraft` 
? 
can_save_draft could have various interpretations.

src/app/cases/creator/case-creator-submit.component.ts
```diff
@@ -44,6 +47,29 @@ export class CaseCreatorSubmitComponent implements OnInit {
...
+    return !this.eventTrigger.draft_store_id;
```
mario-paniccia: can we rename draft_store_id to draft_id?

src/app/cases/creator/case-creator-submit.component.ts
```diff
@@ -44,6 +47,29 @@ export class CaseCreatorSubmitComponent implements OnInit {
...
+  }
+
+  saveDraft(): (caseEventData: CaseEventData) => Observable<Draft> {
+    if (this.canSaveDraft()) {
+      if (this.isCreatingDraft()) {
+        return (caseEventData: CaseEventData) => this.draftService.createDraft(this.jurisdictionId,
```
mario-paniccia: I would have thought the Draft store API would provide an upsert method that creates or updates...do we need to do this check here and call either save or update on the draft store?

src/app/shared/search/search-result.component.ts
```diff
@@ -111,6 +113,12 @@ export class SearchResultComponent implements OnChanges {
     return this.resultView.results.length && this.paginationMetadata.total_pages_count;
   }
 
+  hasDrafts(): boolean {
```
mario-paniccia: same as above. The best place for this method would be on the SearchResultView itself

src/app/shared/search/search-result.component.ts
```diff
@@ -151,4 +159,8 @@ export class SearchResultComponent implements OnChanges {
     }
     return isAscending ? SortOrder.ASCENDING : isDescending ? SortOrder.DESCENDING : SortOrder.UNSORTED;
   }
+
+  removeDraftId(resultViewItem: SearchResultViewItem): string {
```
mario-paniccia: this seems to be used only by the html. Shall we move this to a  pipe?

mario-paniccia: perhaps another solution would be to change the search view to not display the case_id, but a new field `displayed_case_id` of the SearchResultViewItem
for a draft displayed_case_id = 'Draft', for a case it's it's case_id
this way we won't have to mutate the case_id, which could be source of issues

src/app/cases/creator/case-creator-submit.component.ts
mario-paniccia: Have a look at the profile.model.ts, which has a isSolicitor method. 
And the new way we deserialise it in profile.service.ts (i.e. using the library class-transformer)

src/app/cases/creator/case-creator-submit.component.ts
tzarsmango: ok

src/app/shared/case-editor/case-edit-page.component.ts
```diff
@@ -95,23 +95,22 @@ export class CaseEditPageComponent implements OnInit, AfterViewChecked {
...
-        .subscribe(() => this.next(),
-          error => {
-            this.isSubmitting = false;
-            this.error = error;
-            this.callbackErrorsSubject.next(this.error);
-            if (this.error.details) {
-              this.formErrorService
-                .mapFieldErrors(this.error.details.field_errors, this.editForm.controls['data'] as FormGroup, 'validation');
+        .subscribe(() => {
+            if (this.eventTrigger.can_save_draft) {
```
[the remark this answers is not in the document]
tzarsmango: ok

src/app/cases/creator/case-creator-submit.component.ts
```diff
@@ -44,6 +48,25 @@ export class CaseCreatorSubmitComponent implements OnInit {
...
+  saveDraft(): (caseEventData: CaseEventData) => Observable<Draft> {
+    if (this.eventTrigger.can_save_draft) {
+      if (this.isCreatingDraft()) {
+        return (caseEventData: CaseEventData) => this.draftService.createDraft(this.jurisdictionId,
+          this.caseTypeId,
+          caseEventData);
+      } else {
+        return (caseEventData: CaseEventData) => this.draftService.updateDraft(this.jurisdictionId,
+          this.caseTypeId,
+          this.eventTrigger.case_id.slice(CaseResolver.DRAFT.length),
```
mario-paniccia: it's not immediate to figure out what this line is doing. Can we create a method on evenTrigger with a meaningful name ?

src/app/cases/viewer/case-viewer.component.ts
```diff
@@ -105,4 +106,8 @@ export class CaseViewerComponent implements OnInit, OnDestroy {
       return acc.concat(tab.fields);
     }, []);
   }
+
+  isDraft(): boolean {
```
mario-paniccia: Any luck with moving this method into CaseDetails? I still believe that all logic to determine if a case details is a draft would be better placed in one place only

src/app/cases/case.resolver.ts
```diff
@@ -59,16 +62,29 @@ export class CaseResolver implements Resolve<CaseView> {
   }
 
   private getAndCacheCaseView(jid, ctid, cid): Observable<CaseView> {
+    if (cid.startsWith(CaseResolver.DRAFT)) {
```
mario-paniccia: couldn't we have a method on CaseDetails isDraft(cid) ?

src/app/shared/case-editor/case-edit-page.component.ts
```diff
@@ -128,6 +123,9 @@ export class CaseEditPageComponent implements OnInit, AfterViewChecked {
 
   previous(): Promise<boolean> {
     this.error = null;
+    if (this.eventTrigger.can_save_draft) {
```
mario-paniccia: this check, which is duplicated, could be moved inside 
this.saveDraft()

src/app/shared/search/filters/search-filters.component.ts
```diff
@@ -59,9 +60,18 @@ export class SearchFiltersComponent implements OnInit {
   apply(): void {
     this.selected.formGroup = this.formGroup;
     this.selected.page = 1;
+    this.selected.metadataFields = this.getMetadataFields();
     this.onApply.emit(this.selected);
   }
 
+  getMetadataFields(): string[] {
```
mario-paniccia: can we rename to getMetadataFieldsIds?

src/app/shared/search/filters/search-filters.component.ts
```diff
@@ -33,7 +33,8 @@ export class SearchFiltersComponent implements OnInit {
     caseType?: CaseType,
     formGroup?: FormGroup,
     caseState?: CaseState,
-    page?: number
+    page?: number,
+    metadataFields?: string[]
```
mario-paniccia: can we rename to metadataFieldsIds?

src/app/shared/search/search-result-view.model.ts
```diff
@@ -1,7 +1,19 @@
...
   columns: SearchResultViewColumn[];
+
+  @Type(() => SearchResultViewItem)
   results: SearchResultViewItem[];
+  result_error?: string;
+
+  hasDrafts() {
+    return this.results[0]
+    && this.results[0].case_id
+    && this.results[0].case_id.startsWith(CaseResolver.DRAFT);
```
mario-paniccia: this logic is duplicated. We should find a way to centralise it somewhere

mario-paniccia: what did the business say regarding drafts on the search page?

src/app/search/search.component.ts
```diff
@@ -67,6 +84,9 @@ export class SearchComponent implements OnInit {
 
   private getCaseFilterFromFormGroup(formGroup?: FormGroup): object {
     const result = {};
+    result[SearchComponent.METADATA_FILTER] = {};
```
mario-paniccia: feels like things are getting complicated here. 
To simplify a bit the code, can we use types maybe? i.e. introducing  a class which has  2 properties caseFilter and metadataFilter and returning that instead of an Object

src/app/search/search.component.ts
```diff
@@ -35,28 +45,35 @@ export class SearchComponent implements OnInit {
...
     if (filter.caseState) {
       paginationParams['state'] = filter.caseState.id;
       searchParams['state'] = filter.caseState.id;
     }
 
+    if (filter.page) {
+      searchParams['page'] = filter.page;
+    }
+
+    const filters = this.getCaseFilterFromFormGroup(filter.formGroup);
```
mario-paniccia: I think we need to do some refactoring of this search component sooner or later. I feels like things are getting complicated. 
I think all this code to retrieve filters, caseFilters, metadataFilters, metadataPaginationParams, belongs to the filter itself. But I guess we can't easily move it there because filter here is an untyped param :(

src/app/shared/search/filters/search-filters.component.ts
```diff
@@ -33,7 +33,8 @@ export class SearchFiltersComponent implements OnInit {
     caseType?: CaseType,
     formGroup?: FormGroup,
     caseState?: CaseState,
-    page?: number
```
mario-paniccia: sooner or later we have to introduce a proper type here with properties and utility methods

src/app/workbasket/workbasket.component.ts
```diff
@@ -12,6 +12,8 @@ import { CaseType } from '../shared/domain/definition/case-type.model';
 import { FormGroup } from '@angular/forms';
 import { SearchResultComponent } from '../shared/search/search-result.component';
 import { Observable } from 'rxjs/Observable';
+import { AlertService } from '../core/alert/alert.service';
+import { plainToClass } from 'class-transformer';
```
mario-paniccia: this class too could use some refactoring sooner or later

src/app/search/search.component.ts
tzarsmango: this code fragment is from another ticket and got merged here accidentally

src/app/shared/search/filters/search-filters.component.ts
tzarsmango: this code fragment is from another ticket and got merged here accidentally

src/app/workbasket/workbasket.component.ts
tzarsmango: this code fragment is from another ticket and got merged here accidentally

src/app/shared/search/search-result-view.model.ts
tzarsmango: addressed in the next PR https://github.com/hmcts/ccd-case-management-web/pull/183

src/app/shared/case-editor/case-edit-page.component.ts
tzarsmango: addressed in the next [PR](https://github.com/hmcts/ccd-case-management-web/pull/183)

src/app/cases/case.resolver.ts
tzarsmango: addressed in the next [PR](https://github.com/hmcts/ccd-case-management-web/pull/183)

src/app/cases/viewer/case-viewer.component.ts
tzarsmango: We need this since used in the HTML page. Centralisation is addressed in the next [PR](https://github.com/hmcts/ccd-case-management-web/pull/183)

src/app/cases/creator/case-creator-submit.component.ts
tzarsmango: addressed in the next [PR](https://github.com/hmcts/ccd-case-management-web/pull/183)

PR #172, opened 2018-07-16
RDM-2253: Prevent clickjacking

MKaramath:
https://tools.hmcts.net/jira/browse/RDM-2253

### Change description ###
Clickjacking is a technique used by attackers to trick users into executing malicious scripts. This is achieved by embedding a page of the web application on a site controlled by the attacker via an HTML iFrame. Not including X-Frame-Options header leaves the application vulnerable to this attack.

If an attacker is able to embed the page in an HTML iFrame on another site (under the control of the attacker), they can then overlay this page with transparent layers to intercept clicks - causing the victim to inadvertently perform functions or execute malicious code on the transparent page\layer.

```
[ ] Yes
[x] No
```


server.ts
```diff
@@ -57,6 +57,12 @@ app.engine('html', ngExpressEngine({
 app.set('view engine', 'html');
 app.set('views', join(DIST_FOLDER, 'browser'));
 
+let xFrameOptions = require('x-frame-options');
```
vlaurin: `import` should be used instead of `require`; and this should be placed at the top of the files with the other imports.

server.ts
```diff
@@ -57,6 +57,12 @@ app.engine('html', ngExpressEngine({
...
+app.use(xFrameOptions());
+app.get('/', function (req, res) {
+  res.setHeader('X-Frame-Options', 'DENY');
+});
```
vlaurin: I believe the `app.get()` part above can be removed as it's duplicating the implementation of `xFrameOptions`, see https://github.com/domharrington/x-frame-options/blob/master/index.js#L5

server.ts
MKaramath: Done.

PR #174, opened 2018-07-19
RDM-900 - Allow any case field to be first column of case list or search results

gbenadikar:
https://tools.hmcts.net/jira/browse/RDM-900
https://tools.hmcts.net/jira/browse/RDM-1508

### Change description ###
This change serves both the stories. Regardless of the field type either flexible field or metadata,  first column will be shown with hyperlink. When a case does not have value for the first column, default value shown will be case reference (regardless of the column header)

```
[x] Yes
[] No
```
The definitions files will have to be changed to include [CASE_REFERENCE] as the first column in work basket results and search results to retain current functionality with no surprises to the users.


src/app/shared/search/search-result.html
```diff
@@ -17,11 +16,18 @@
...
-    <td *ngFor="let col of resultView.columns">
-      <div class="text-16" [style.visibility]="hideRows ? 'hidden' : 'visible'">
+    <td *ngFor="let col of resultView.columns; let colIndex = index">
+      <a *ngIf="colIndex == 0" routerLink="/case/{{jurisdiction.id}}/{{caseType.id}}/{{result.case_id}}">
+        <div class="text-16" [style.visibility]="hideRows ? 'hidden' : 'visible'">
+          <ccd-field-read [caseField]="{
+                  id: col.case_field_id,
+                  label: col.label,
+                  field_type: col.case_field_type,
+                  value: result.case_fields[col.case_field_id]? col.case_field_id === 'case_reference' ? (result.case_fields[col.case_field_id] | ccdCaseReference) : result.case_fields[col.case_field_id] : (result.case_id | ccdCaseReference)
```
vlaurin: I fear this won't work if the first column's field type is not Text or Number (e.g. fixed list, document, moneyGBP, ...).

I would recommend replacing the ternary conditions on `ccd-field-read`'s value, by an if/else block with different templates as this would resolve the above concern.

gbenadikar: Thanks. I have corrected the code as discussed using templates. Please re-review.

PR #179, opened 2018-07-23
RDM-2389 fixed a bug in mandatory field check 

tzarsmango:
fixed a bug in mandatory field check for document field type with conditional hide'n'show

https://tools.hmcts.net/jira/browse/RDM-2389

```
[ ] Yes
[X] No
```

### DIFF
--- a/src/app/shared/case-editor/case-edit-page.component.ts
+++ b/src/app/shared/case-editor/case-edit-page.component.ts
@@ -13,6 +13,7 @@ import { FormErrorService } from '../../core/form/form-error.service';
 import { CallbackErrorsContext } from '../error/error-context';
 import { CaseFieldService } from '../domain/case-field.service';
 import { CaseField } from '../domain/definition/case-field.model';
+import { ShowCondition } from '../conditional-show/conditional-show.model';
 
 @Component({
   selector: 'ccd-case-edit-page',
@@ -72,7 +73,8 @@ export class CaseEditPageComponent implements OnInit, AfterViewChecked {
       .filter(caseField => !this.caseFieldService.isReadOnly(caseField))
       .every(caseField => {
         let theControl = this.editForm.controls['data'].get(caseField.id);
-        return this.checkDocumentField(caseField, theControl) && this.checkOptionalField(caseField, theControl);
+        return this.isHidden(caseField, this.editForm.getRawValue())
+          || (this.checkDocumentField(caseField, theControl) && this.checkOptionalField(caseField, theControl));
       });
   }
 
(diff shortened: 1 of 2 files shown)


src/app/shared/case-editor/case-edit-page.component.ts
```diff
@@ -72,7 +73,8 @@ export class CaseEditPageComponent implements OnInit, AfterViewChecked {
       .filter(caseField => !this.caseFieldService.isReadOnly(caseField))
       .every(caseField => {
         let theControl = this.editForm.controls['data'].get(caseField.id);
-        return this.checkDocumentField(caseField, theControl) && this.checkOptionalField(caseField, theControl);
+        return this.isHidden(caseField, this.editForm.getRawValue())
```
mario-paniccia: there's a lot of complexity growing here. If you have time @tzarsmango, it would be good to move all the logic of the method currentPageIsNotValid into a separate service PageValidationService

mario-paniccia: maybe it's a bit simpler to read if we move this isHidden check as another filter after 
```.filter(caseField => !this.caseFieldService.isReadOnly(caseField))```

**Comment by mario-paniccia** [comment]
@vlaurin 
good point. Checking whether the field is already disabled/hidden rather than re evaluating the show condition might work. It will also avoid introducing here a dependency on ShowCondition. Worth giving it a try

File: src/app/shared/case-editor/case-edit-page.component.ts (hunk 1 above)
Comment (tzarsmango) [reply src/app/shared/case-editor/case-edit-page.component.ts]: yes, absolutely better. Thanks for pointing out

Comment (tzarsmango) [reply src/app/shared/case-editor/case-edit-page.component.ts]: 👍

**Comment by tzarsmango** [comment]
Addressed most of the comments, left one to the following tech debt task: https://tools.hmcts.net/jira/browse/RDM-2539

File: src/app/shared/case-editor/page-validation.service.ts [hunk 2]
```diff
@@ -0,0 +1,41 @@
...
+import { CaseFieldService } from '../domain/case-field.service';
+import { WizardPage } from '../domain/wizard-page.model';
+
+@Injectable()
+export class PageValidationService {
+  constructor(private caseFieldService: CaseFieldService) {}
+
+  isPageValid(page: WizardPage, editForm: FormGroup): boolean {
+    return page.case_fields
+      .filter(caseField => !this.caseFieldService.isReadOnly(caseField))
```
mario-paniccia: Fatih can you please add a TODO comment inside CaseFieldService to get rid of CaseFieldService and move all its methods inside CaseField itself please?

PR #181, opened 2018-07-25
RDM-2432: Document upload - Chosen file name is not retained

MKaramath:
https://tools.hmcts.net/jira/browse/RDM-2432

### Change description ###
Document upload: Chosen file name is not retained when user navigates back to the previous page.

Steps to repro:

ChooStylese a file by clicking browse button, file name appears next to 'Browse' button
Click Continue to navigate to next page
Again click Previous button on the next page to navigate back to the previous page
Expected: Chosen file name should appear in the 'Document URL' next to Browse button

Actual : Chosen file name disappear from the Document URL box. However, the document gets uploaded when the case/event is submitted.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/app/shared/palette/document/write-document-field.component.ts
+++ b/src/app/shared/palette/document/write-document-field.component.ts
@@ -37,7 +37,6 @@ export class WriteDocumentFieldComponent extends AbstractFieldWriteComponent imp
       let documentUpload: FormData = new FormData();
       documentUpload.append('files', this.selectedFile, this.selectedFile.name);
       documentUpload.append('classification', 'PUBLIC');
-      this.caseField.value = this.selectedFile.name;
       this.documentManagement.uploadFile(documentUpload).subscribe(result => {
         if (!this.uploadedDocument) {
           this.createDocumentGroup();
@@ -49,7 +48,7 @@ export class WriteDocumentFieldComponent extends AbstractFieldWriteComponent imp
           document._links.binary.href,
           document.originalDocumentName,
         );
-
+        this.caseField.value = this.selectedFile.name;
         this.valid = true;
       }, (error: HttpError) => {
         this.uploadError = this.getErrorMessage(error);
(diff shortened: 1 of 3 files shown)


src/app/shared/palette/document/write-document-field.component.ts
```diff
@@ -48,7 +48,7 @@ export class WriteDocumentFieldComponent extends AbstractFieldWriteComponent imp
           document._links.binary.href,
           document.originalDocumentName,
         );
-
+        this.caseField.value = this.selectedFile.name;
```
mario-paniccia: shouldn't we also set the caseField.value to null later on the way we set 
`      this.selectedFile = null;
`
on line 58?

mario-paniccia: also, aren't we using this.caseField.value already with a different meaning? on the ngOnInit:
```ngOnInit() {
    let document = this.caseField.value;
    if (document) {
      this.createDocumentGroup(
        document.document_url,
        document.document_binary_url,
        document.document_filename,
      );
    }
  }

File: src/app/shared/palette/document/write-document-field.html [hunk 2]
```diff
@@ -8,6 +8,7 @@
 
   <div>
     <ccd-read-document-field *ngIf="caseField" [caseField]="caseField"></ccd-read-document-field>
+    <div class="chosen-file" *ngIf="selectedFile && selectedFile.name && caseField.value">Chosen file: {{ caseField.value }}</div>
```
mario-paniccia: can we simplify this to:
`<div class="chosen-file" *ngIf="selectedFile">Chosen file: {{ selectedFile.name }}</div>`
If not, can we move this long if condition inside a method in the component called something like 'isFileUploaded()'

MKaramath:
Closing this PR, since this change will be fixed as part of RDM-1124.

PR #196, opened 2018-07-31
RDM-1289 - make case title configurable

gbenadikar:
https://tools.hmcts.net/jira/browse/RDM-1289

### Change description ###
Code changes to show custom title when defined. Custom title is a markdown element with label substitution (label interpolation). Custom title will be displayed only if defined in the state tab of the definition file.

```
[ ] Yes
[x] No
```


src/app/shared/header/case-header.component.spec.ts
```diff
@@ -39,25 +30,39 @@ describe('CaseHeaderComponent', () => {
...
     expect(text(header)).toEqual('#1234-5678-9012-3456');
   });
 
+  it('should render a header with markdown element when title display is not empty', () => {
+    component.caseDetails.state.title_display = 'Title';
+    component.ngOnInit();
+    fixture.detectChanges();
+
+    let header = de.query($MARKDOWN);
+    expect(header).toBeTruthy();
```
vlaurin: Could we make it so that the actual content of the title, in this case "Title", is tested as well?

src/app/shared/header/case-header.component.ts
```diff
@@ -1,13 +1,32 @@
...
+      this.caseFields = this.getCaseFields();
+    }
+  }
+
+  private getCaseFields(): CaseField[] {
+    const caseDataFields = this.caseDetails.tabs.reduce((acc, tab) => {
+      return acc.concat(tab.fields);
+    }, []);
+
+    return Object.assign(caseDataFields, this.caseDetails.metadataFields);
```
vlaurin: Shouldn't that be using `Array#concat()` instead? :)

mario-paniccia: how come we are using Object.assign? You just want to add the metadata fields to the array am I correct? If yes, would this work:
` caseDataFields.concat(this.caseDetails.metadataFields);`

src/app/shared/header/case-header.component.ts
```diff
@@ -1,13 +1,32 @@
...
 
+  ngOnInit(): void {
+    this.caseTitle = new CaseField();
+    if (this.caseDetails.state.title_display) {
+      this.caseTitle.label = this.caseDetails.state.title_display;
+      this.caseFields = this.getCaseFields();
+    }
+  }
+
+  private getCaseFields(): CaseField[] {
```
mario-paniccia: I think this method belongs to CaseView itself. Can we move it there?

src/app/shared/header/case-header.html
```diff
@@ -1,3 +1,8 @@
-<h2 class="heading-large">
+<h2 *ngIf="!caseTitle.label; else case_title" class="heading-large">
```
mario-paniccia: 1) can we make this more self explanatory introducing a boolean method in the component: 
`displayCustomTitle()`
2) can we avoid using else inside the ngIf? I think this reads better:
```
<h2 *ngIf="!displayCustomTitle()" ...>
...
<ng-template *ngIf="displayCustomTitle()"...>

File: src/app/shared/header/case-header.component.spec.ts (hunk 1 above)
Comment (gbenadikar) [reply src/app/shared/header/case-header.component.spec.ts]: Done

File: src/app/shared/header/case-header.component.ts (hunk 2 above)
Comment (gbenadikar) [reply src/app/shared/header/case-header.component.ts]: Of course

Comment (gbenadikar) [reply src/app/shared/header/case-header.component.ts]: :-)

Comment (gbenadikar) [reply src/app/shared/header/case-header.component.ts]: yeah rectified.

File: src/app/shared/header/case-header.component.ts (hunk 3 above)
Comment (gbenadikar) [reply src/app/shared/header/case-header.component.ts]: Tried moving but didn't work. 'Not a function' error. Not a straight forward one, so will leave it out.

File: src/app/shared/header/case-header.html (hunk 4 above)
Comment (gbenadikar) [reply src/app/shared/header/case-header.html]: Changed to make it more readable

File: src/app/shared/header/case-header.html [hunk 5]
```diff
@@ -1,3 +1,8 @@
...
   #{{ caseDetails.case_id | ccdCaseReference}}
 </h2>
+<ng-template #case_title ccdLabelSubstitutor [caseField]="caseTitle" [eventFields]="caseFields">
```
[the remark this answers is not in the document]
gbenadikar: Yes modified label field component, please review carefully :)

src/app/shared/header/case-header.component.ts
gbenadikar: Will handle this as part of https://tools.hmcts.net/jira/browse/RDM-2681

PR #208, opened 2018-08-07
[PREVIEW] RDM-2348: Grey bar on conditionally show/hide fields

MKaramath:
https://tools.hmcts.net/jira/browse/RDM-2348

### Change description ###
The idea of this change is when a field on a current page - hide/show's another field or fields on the same page - a grey vertical bar on the left should be show.

```
[ ] Yes
[x] No
```


src/app/shared/conditional-show/conditional-show.directive.ts
```diff
@@ -100,6 +100,13 @@ export class ConditionalShowDirective implements AfterViewInit, OnDestroy {
 
   private show() {
     this.el.nativeElement.hidden = false;
+    if (this.caseField && this.caseField.field_type && this.caseField.field_type.type !== 'Collection') {
```
mario-paniccia: is `this.caseField && this.caseField.field_type` needed? 
if this.caseField was null, other parts of this class would fail. Also, the property `field_type` is mandatory. 
I'm not 100% but can you try to remove `this.caseField && this.caseField.field_type` and see if it still works?

mario-paniccia: we should move this whole new block of code into it's own private method, with a descriptive name to make the code more understandable. 
I guess we could call this new method `showGrayBar` ?

mario-paniccia: I think the conditional-show.directive should not know the details of how to display grey bar. The directive should only have logic to decide if a field has to be displayed or not. 
Is it possible to move the method showGrayBar() outside of this class, for example having it on the field itself, and just call it here?

src/app/shared/conditional-show/conditional-show.directive.spec.ts
```diff
@@ -66,6 +67,20 @@ describe('ConditionalShowDirective', () => {
         expect(conditionalShow.condition).toBeUndefined();
     });
```
mario-paniccia: the story mentions that the grey bar has to be displayed only if the show is triggered from field on the current page. 
If the field is displayed from fields of another page, grey bar should not be displayed. Can we have a test for this?

src/app/shared/conditional-show/conditional-show.directive.ts
```diff
@@ -100,6 +100,13 @@ export class ConditionalShowDirective implements AfterViewInit, OnDestroy {
...
+      if (this.el.nativeElement.closest('ccd-write-complex-type-field')) {
+        this.el.nativeElement.children[0].children[0].classList.add('show-condition-gray-bar');
```
mario-paniccia: this seems a bit fragile to me

src/app/shared/conditional-show/conditional-show.directive.ts
```diff
@@ -100,6 +100,13 @@ export class ConditionalShowDirective implements AfterViewInit, OnDestroy {
 
   private show() {
     this.el.nativeElement.hidden = false;
+    if (this.caseField && this.caseField.field_type && this.caseField.field_type.type !== 'Collection') {
+      if (this.el.nativeElement.closest('ccd-write-complex-type-field')) {
```
mario-paniccia: this way of testing for a complex type is a bit fragile

src/app/shared/conditional-show/conditional-show.directive.ts
```diff
@@ -100,6 +100,13 @@ export class ConditionalShowDirective implements AfterViewInit, OnDestroy {
...
+        this.el.nativeElement.children[0].children[0].classList.add('show-condition-gray-bar');
+      } else {
+        this.el.nativeElement.classList.add('show-condition-gray-bar');
+      }
+    }
```
vlaurin: The intent of the whole code block above is lost in the complexity of the conditions and deep object navigation I think. Introducing private method and/or intermediary variables would make it more meaningful.

src/style/app.scss
```diff
@@ -226,9 +226,14 @@ select::-ms-expand {
...
   margin-bottom: 25px;
 }
-
 #fileInputWrapper{
   position:absolute;
   width:50%;
   height:41px;
 }
+
+.show-condition-gray-bar {
```
vlaurin: Couldn't the Gov.UK class for the gray bar be re-used instead of introducing a new class?

src/app/shared/conditional-show/conditional-show.directive.ts
MKaramath: field_type is mandatory, but in it is failing most of the test case mocks so to be on safer side, I have added this check.

src/style/app.scss
MKaramath: Done.

src/app/shared/conditional-show/conditional-show.directive.ts
MKaramath: Done.

src/app/shared/conditional-show/conditional-show.directive.ts
MKaramath: Done.

src/app/shared/conditional-show/conditional-show.directive.ts
MKaramath: Done.

src/app/shared/conditional-show/conditional-show.directive.ts
MKaramath: Done.

src/app/shared/utils/fields.utils.ts
```diff
@@ -122,4 +122,13 @@ export class FieldsUtils {
     return Object.assign({}, obj);
   }
 
+  showGrayBar(caseField: CaseField, element: any) {
```
vlaurin: Instead of `any`, I believe the element should be typed as `HTMLElement`.

MKaramath: Done.

Saravanabh:
Completed testing and this is now ready to be merged.

src/app/shared/utils/fields.utils.ts
```diff
@@ -122,4 +122,24 @@ export class FieldsUtils {
     return Object.assign({}, obj);
   }
 
+  showGrayBar(caseField: CaseField, element: HTMLElement) {
+    if (caseField && caseField.field_type && caseField.field_type.type !== 'Collection') {
```
mario-paniccia: caseField check is not needed, remove please. In the code we are always passing a value for case field. And if that didnt' happen, the conditional show directive will break even before this method is called. 

Also caseField.field_type check is not required because fieldType is mandatory. If this breaks the tests, then the test have to be updated rather that polluting the production code with unnecessary null checks all over the place

MKaramath: Done.  Removed those checks.   I had added them because it was failing several unit tests.  Now I fixed all the test cases.  All fine now.

package.json
```diff
@@ -18,7 +18,7 @@
     "test:a11y": "echo 'TODO: Accessibility tests'",
     "test:nsp": "nsp check",
     "test:cucumber": "protractor test/config/cucumber.conf.js",
-    "test:smoke": "protractor test/config/smoke.conf.js",
+    "test:smoke": "echo 'TODO: Enable Smoke tests'",
```
mario-paniccia: have we disabled smoke tests?

src/app/shared/conditional-show/conditional-show.directive.spec.ts
```diff
@@ -66,6 +74,22 @@ describe('ConditionalShowDirective', () => {
...
 
+    it('should display gray bar when the field is on the same page', () => {
+      comp.caseField = field('PersonSecondAddress', '', 'PersonLastName="Doe"');
+      expect(comp.caseField.field_type.type).toBe('Text');
+      comp.eventFields = [comp.caseField, field('PersonLastName', 'Doe', '')];
+      fixture.detectChanges();
+      expect(el.hidden).toBe(false);
+    });
+
+    it('should display gray bar when the field is not on the same page', () => {
```
mario-paniccia: should not display?

src/app/shared/conditional-show/conditional-show.directive.spec.ts
```diff
@@ -66,6 +74,22 @@ describe('ConditionalShowDirective', () => {
...
+      comp.caseField = field('PersonSecondAddress', '', 'PersonLastName="Doe"');
+      expect(comp.caseField.field_type.type).toBe('Text');
+      comp.eventFields = [comp.caseField, field('OtherField', 'Doe', '')];
+      fixture.detectChanges();
+      expect(el.hidden).toBe(true);
+    });
+
     it('should display when condition matches a read only field. No form fields', () => {
```
mario-paniccia: there's no test that verifies this condition:
```if (caseField.field_type.type !== 'Collection') {```

File: src/app/shared/utils/fields.utils.ts [hunk 12]
```diff
@@ -122,4 +122,35 @@ export class FieldsUtils {
...
+        }
+        if (!cyaSelector && samePage) {
+          divSelector.classList.add('panel');
+        }
+      }
+    }
+  }
+
+  getConditionFields(condition: string): any[] {
+    let condFields = [];
```
mario-paniccia: we should not be parsing a show condition in fields.utils.ts. This violates separation of concerns principle. fields.utils.ts should know nothing about how to parse a show condition and extract his fields. 
This knowledge is encapsulated in our class ShowCondition. 
This logic should be moved there and unit tests should be added to `conditional-show.model.spec.ts`

src/app/shared/utils/fields.utils.ts
```diff
@@ -122,4 +122,35 @@ export class FieldsUtils {
     return Object.assign({}, obj);
   }
 
+  showGrayBar(caseField: CaseField, element: HTMLElement) {
```
mario-paniccia: if you look at the rest of methods in fields.utils.ts, you'll see that this class is not concerned with UI modifications. So a method like `showGrayBar` does not belong here.
I was thinking of introducing a new class display.utils.ts and move it there. But perhaps we can find some other placement for this method which is even better, I'll give it a thought and come back to you

package.json
vlaurin: This could be a matter of this branch being out-of-sync with the `master` branch.
@MKaramath could you please merge `master` in this branch and double-check that this pull request doesn't contains any changes on `package.json`

src/app/shared/utils/fields.utils.ts
```diff
@@ -122,4 +122,35 @@ export class FieldsUtils {
...
+    if (caseField.field_type.type !== 'Collection') {
+      let divSelector = element.querySelector('div');
+      let samePage = true;
+      let cyaSelector = false;
+      this.getConditionFields(caseField.show_condition).forEach(
+        field => { samePage = document.querySelector('#' + field) ? true : false; }
+      );
+      if (divSelector) {
+        let tempElement = divSelector.parentElement;
```
mario-paniccia: I think using dom navigation makes this method a bit fragile. It's not ideal. Let see if we can think together of a better solution. 
I'll give it a thought

mario-paniccia:
replaced by RDM-2348_1

PR #212, opened 2018-08-08
[PREVIEW] RDM-2493: Date/Time should be displayed in local time zone

MKaramath:
https://tools.hmcts.net/jira/browse/RDM-2493

### Change description ###
The event created date/time is saved in UTC however it  also displayed in UTC. This should instead be displayed in BST.

```
[ ] Yes
[x] No
```


src/app/shared/palette/utils/date.pipe.spec.ts
```diff
@@ -3,99 +3,105 @@ import { DatePipe } from './date.pipe';
 describe('DatePipe', () => {
 
   let datePipe: DatePipe;
-  // const EXPECTED_OFFSET = - new Date().getTimezoneOffset() / 60;
-  const EXPECTED_OFFSET = 0; // RDM-1149 changed the pipe logic
+  const EXPECTED_OFFSET = - new Date().getTimezoneOffset() / 60;
+  // const EXPECTED_OFFSET = 0; // RDM-1149 changed the pipe logic
```
adr1ancho: Line 7 is not required

src/app/shared/palette/utils/date.pipe.ts
```diff
@@ -12,13 +12,19 @@ export class DatePipe implements PipeTransform {
     ['Jan'], ['Feb'], ['Mar'], ['Apr'], ['May'], ['Jun'], ['Jul'], ['Aug'], ['Sep'], ['Oct'], ['Nov'], ['Dec'],
   ];
 
-  transform(value: string, format: string): string {
+  transform(value: string, zone: string, format: string): string {
```
adr1ancho: Can we have one more function

transform(value: string, format: string): string {
    return transform(value, format, null)
}

MKaramath: As discussed, the method overloading is not supporting in angular.

src/app/shared/palette/utils/date.pipe.spec.ts
MKaramath: Done.

src/app/shared/palette/utils/date.pipe.ts
adr1ancho: thx

PR #215, opened 2018-08-09
Nightly: build on the nightly pipeline

dwaynebailey:
https://tools.hmcts.net/jira/browse/RDM-2311

```
[ ] Yes
[ ] No
```

### DIFF
--- a/Jenkinsfile_nightly
+++ b/Jenkinsfile_nightly
@@ -1,9 +1,9 @@
 #!groovy
 
-properties(
+properties([
     // H allow predefined but random minute see https://en.wikipedia.org/wiki/Cron#Non-standard_characters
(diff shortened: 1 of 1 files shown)


Jenkinsfile_nightly
```diff
@@ -1,16 +1,16 @@
...
+])
 
 @Library("Infrastructure")
 
 def type = "angular"
 def product = "ccd"
 def app = "case-management-web"
 
-withPipeline(type, product, app) {
+withNightlyPipeline(type, product, component) {
```
vlaurin: `app` was correct, as defined above no? `component` would be undefined

dwaynebailey: Not having a great day with that one!

PR #253, opened 2018-09-05
RDM-2757: Display number of cases and total number

MKaramath:
https://tools.hmcts.net/jira/browse/RDM-2757

### Change description ###
Display the number of cases and the total number on the top of the pagination results section 

```
[ ] Yes
[x] No
```


src/app/search/search.component.html
```diff
@@ -1,4 +1,4 @@
-<h3 class="heading-medium">Search</h3>
+<h3 class="heading-medium" style="margin-top: 25px;">Search</h3>
```
mohammedanas: It could be nice to have it as a class in scss and added it in the class eg:-class= "heading-medium search-margin" instead of margin-top: 25px;

src/app/shared/search/search-result.component.ts
```diff
@@ -157,4 +157,14 @@ export class SearchResultComponent implements OnChanges {
     }
     return isAscending ? SortOrder.ASCENDING : isDescending ? SortOrder.DESCENDING : SortOrder.UNSORTED;
   }
+
+  getFirstResult(): number {
+    let currentPage = (this.selected.page ? this.selected.page : 1);
```
mohammedanas: const would be a better option instead of let as this is not modified with in the scope

src/app/shared/search/search-result.component.ts
```diff
@@ -157,4 +157,14 @@ export class SearchResultComponent implements OnChanges {
...
+    return ( (currentPage - 1) * this.paginationPageSize ) + 1
+  }
+
+  getLastResult(): number {
+    let currentPage = (this.selected.page ? this.selected.page : 1);
```
mohammedanas: const would be a better option instead of let as this is not modified with in the scope

MKaramath: Done.

src/app/shared/search/search-result.component.ts
MKaramath: Done.

src/app/search/search.component.html
MKaramath: Done.

Saravanabh:
Tested in preview and looks good. Ready to be merged.

PR #278, opened 2018-09-17
[PREVIEW] Search filter selects the first case-type & removed unused imports

tzarsmango:
Altered search filters behaviour. Now 1st case-type in the list is selected by default

```
[ ] Yes
[X] No
```


src/app/shared/search/filters/search-filters.component.ts
```diff
@@ -108,11 +106,4 @@ export class SearchFiltersComponent implements OnInit {
     return this.selected.jurisdiction === null  ||
            this.selected.jurisdiction === undefined;
   }
-
-  private selectCaseType(caseTypes: CaseType[]) {
-    if (caseTypes.length === 1) {
-      this.selected.caseType = caseTypes[0];
-      this.onCaseTypeIdChange();
```
andrewfolga: do we call this one now?

tzarsmango: nope, we select the first case type in the list as default behaviour

PR #307, opened 2018-09-28
Revert "RDM-3067: Change IDAM FE DNS to platform.hmcts.net"

vlaurin:
Reverts hmcts/ccd-case-management-web#305


vlaurin:
DNS issue was addressed, revert is not needed anymore.

PR #316, opened 2018-10-03
[PREVIEW] RDM-2901: Significant item link frontend change

MKaramath:
https://tools.hmcts.net/jira/browse/RDM-2901

### Change description ###
If significant item link or document has been attached during an event, then event logs should display the icon with the corresponding link and description in the history tab

```
[ ] Yes
[x] No
```


src/app/cases/viewer/case-viewer.component.spec.ts
```diff
@@ -124,7 +124,12 @@ describe('CaseViewerComponent', () => {
       state_name: 'Case Updated',
       user_id: 0,
       user_last_name: 'Chan',
-      user_first_name: 'Phillip'
+      user_first_name: 'Phillip',
+      significant_item: {
```
tzarsmango: shouldn't this be optional?

src/app/core/cases/case-view-event.model.ts
```diff
@@ -10,4 +10,9 @@ export class CaseViewEvent {
   user_id: number;
   user_last_name: string;
   user_first_name: string;
+  significant_item: {
```
tzarsmango: if we make significant item optional we won't need this and the below additions to the CaseView object...

PR #354, opened 2018-10-18
[PREVIEW] Rdm 3178 + RDM 3179 + bug fix to cancel on CYA page of create case event with S&R: Release new version

andrewfolga:
https://tools.hmcts.net/jira/browse/RDM-3178
https://tools.hmcts.net/jira/browse/RDM-3179

### Change description ###
'Cancel and Return' functionality is enabled when Save and Resume is not configured
As a user I want to 'Cancel and Return' only on a create event
bug fix to cancel on CYA page of create case event with S&R: Release new version

```
[ ] Yes
[x] No
```


src/app/cases/creator/case-creator-submit.component.spec.ts
```diff
@@ -282,5 +284,165 @@ describe('CaseCreatorSubmitComponent', () => {
     expect(draftService.createOrUpdateDraft).toHaveBeenCalledWith(JID, CTID, EVENT_TRIGGER.case_id, SANITISED_EDIT_FORM);
     expect(router.navigate).toHaveBeenCalledWith([`case/${JID}/${CTID}/${EVENT_TRIGGER.case_id}`]);
   });
+});
+
+describe('CaseCreatorSubmitComponent with Save and Resume enabled', () => {
```
andrewfolga: This should end ...disabled instead of ...enabled but let's not raise it here. I will fix it in next PR 3064

PR #366, opened 2018-10-19
Rdm 3064 Case progression: Better component contract

andrewfolga:
https://tools.hmcts.net/jira/browse/RDM-3064

### Change description ###
Case progression: Better component contract:

Removal of CasesService and DraftService as moved to UI toolkit in [105](https://github.com/hmcts/ccd-case-ui-toolkit/pull/105)
Facilitate configuration of case progression component

```
[ ] Yes
[x] No But needs to be release with correct UI toolkit version which at the moment is 1.2.22
```


README.md
```diff
@@ -19,7 +19,7 @@ $ yarn stub-api
 # start the development server
 $ yarn start
 ```
-go to [http://localhost:3451](http://localhost:3451) in your browser.
+go to [http://localhost:3451](http://localhost:3451) in your browser.
```
vlaurin: I'm not sure this change is needed 🤡

src/app/cases/cases.module.ts
```diff
@@ -49,6 +50,8 @@ import { BrowserAnimationsModule } from '@angular/platform-browser/animations';
     CaseEditPageComponent,
     CaseEditFormComponent,
     CaseEditSubmitComponent,
+    CaseCreateComponent,
+    CaseProgressComponent,
```
vlaurin: Why are those 2 declarations required given that they are not used in this app?

README.md
andrewfolga: This was to release in PREVIEW.

src/app/cases/cases.module.ts
andrewfolga: Looks like accidental check in - will clean up.

README.md
vlaurin: You can push empty commit for that :)

PR #377, opened 2018-10-31
Rdm 2323 - Change related to label substitution

mohammedanas:
https://tools.hmcts.net/jira/browse/RDM-2323
User Story
Full Name

As a court staff user,
When I view WorkBasketResults or SearchResults screens
I expect to see the citizens first name and last name joined together as a well formed full name.
And the full name is displayed in a single column.

```
[ ] Yes
[X] No
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,7 +52,7 @@
     "@angular/material": "^6.3.0",
     "@angular/platform-browser": "~6.1.8",
     "@angular/platform-browser-dynamic": "~6.1.8",
-    "@angular/platform-server": "~6.1.8",
+    "@angular/platform-server": "~  6.1.8",
     "@angular/router": "~6.1.8",
     "@angular/upgrade": "~6.1.8",
     "@hmcts/ccd-case-ui-toolkit": "1.2.21",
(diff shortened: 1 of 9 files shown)


src/app/shared/search/search-result.component.ts
```diff
@@ -144,14 +199,16 @@ export class SearchResultComponent implements OnChanges {
     return this.activityService.isEnabled;
   }
 
-  hyphenateIfCaseReferenceOrGet(col, result): any {
-    return col.case_field_id === '[CASE_REFERENCE]' ?
-      this.caseReferencePipe.transform(result.case_fields[col.case_field_id])
-      : result.case_fields[col.case_field_id];
+  getValueFromCaseField(col, result): any {
```
gbenadikar: If a [CASE_REFERENCE] field is used in search results column, it will be expected with hyphens. Is this change not removing that feature?

package.json
```diff
@@ -52,7 +52,7 @@
     "@angular/material": "^6.3.0",
     "@angular/platform-browser": "~6.1.8",
     "@angular/platform-browser-dynamic": "~6.1.8",
-    "@angular/platform-server": "~6.1.8",
+    "@angular/platform-server": "~  6.1.8",
```
gbenadikar: space?

src/app/shared/search/search-result.component.ts
mohammedanas: I have moved this change in case tool kit

PR #382, opened 2018-10-31
[PREVIEW] RDM-2886: Making column sort link accessible

MKaramath:
https://tools.hmcts.net/jira/browse/RDM-2886

### Change description ###
Making column sort link accessible and fixing the existing bug while sorting the column

```
[ ] Yes
[x] No
```


src/app/shared/search/search-result.html
```diff
@@ -5,7 +5,8 @@
   <thead>
   <tr>
     <th *ngFor="let col of resultView.columns">
-      <table class="search-result-column-header">
+      <table class="search-result-column-header"
+             attr.aria-label="Sort by {{col.label}} {{ isSortAscending(col)? 'ascending' : 'descending' }}">
```
mario-paniccia: can we find a way to avoid the UI breaking if we forget to add a new type to the search-result-view-item-comparator-factory.ts? 
Could we catch the example for example and log a warning message?

MKaramath: Fixed this issue.  Please code review.

src/app/shared/search/search-result.component.ts
```diff
@@ -164,6 +166,11 @@ export class SearchResultComponent implements OnChanges {
     let isAscending = true;
     let isDescending = true;
 
+    if (this.comparator(column) === undefined) {
+      console.error('Error: The case_field_type ' + column.case_field_type.type + ' cannot be sorted on ' +
```
mario-paniccia: can we log this as a `warning` rather than an error? Also, can we change the message? e.g.
console.warn('Sort error: unknown sort comparator for ' + column.case_field_type.type)

src/app/shared/search/search-result.component.ts
```diff
@@ -128,7 +128,9 @@ export class SearchResultComponent implements OnChanges {
   }
 
   sort(column: SearchResultViewColumn) {
```
mario-paniccia: where is the test for the case when the 
```this.comparator(column) === undefined```
?

src/app/shared/search/search-result.component.ts
MKaramath: Done.

src/app/shared/search/search-result.component.ts
MKaramath: Done.

PR #416, opened 2018-11-14
added scenarios for case progression

noronhaa:
### JIRA link (if applicable) ###
https://tools.hmcts.net/jira/browse/RDM-3228

### Change description ###
Add functional test for case progression 

**Does this PR introduce a breaking change?** (check one with "x")

```
[ ] Yes
[X] No
```

PR #427, opened 2018-11-16
Revert "change smoke test run format to old original smoke tests"

hemantt:
Reverts hmcts/ccd-case-management-web#423


noronhaa:
Have changed smoke tests back to cucumber test format in #416

PR #517, opened 2019-02-07
RDM-3675 Remove the unused image from public resources. It is now sou…

andrewfolga:
…rced from the lib component.

### Change description ###
A fix to image sourcing. It has to be defined as close to the component as possible and not in the library clients.

```
[ ] Yes
[x] No
```


noronhaa:
covered in #525

PR #523, opened 2019-02-08
Use ACR cache optimisation in CI

slyg:
### JIRA link ###

https://tools.hmcts.net/jira/browse/CNP-1131

### Change description ###

This PR leverages on layers re-use to reduce the the docker image build time in CI.

The image build is **8 min** shorter.

Does this PR introduce a breaking change? (check one with "x")

[] Yes
[x] No


.dockerignore
```diff
@@ -8,3 +8,5 @@
 !tsconfig.json
 !webpack.server.config.js
 !server.ts
+!Dockerfile
```
slyg: Because ACR is a bit silly sometimes

PR #550, opened 2019-02-21
RDM-3699: Fonts improvement through out CCD

MKaramath:
https://tools.hmcts.net/jira/browse/RDM-3699

### Change description ###
Fonts improvement through out CCD

Related PR in ccd-case-ui-toolkit:
https://github.com/hmcts/ccd-case-ui-toolkit/pull/231

```
[ ] Yes
[x] No
```


src/app/footer-nav/cookies.component.html
```diff
@@ -12,7 +12,7 @@ <h1 class="heading-large">Cookies</h1>
           when you visit a website. Most websites need cookies to work properly.</span>
       </div>
 
-      <h3 class="heading-medium">How cookies are used in this service</h3>
+      <h3 class="heading-h3">How cookies are used in this service</h3>
```
mario-paniccia: Can't we create a common class, say .cookies-header-size, set it up with a heading-h3, and use that one here? that way it'll be simpler changing it in the future if needed

MKaramath: Since I am using standard h3 header across the CCD including cookies, I am using same class here and in the rest of the places as well

veeresha-hmcts:
Verified on AAT and PR approved

PR #553, opened 2019-02-22
update vault keys for jenkins pipeline

noronhaa:
n/a

### Change description ###
update vault keys for new user for functional tests

```
[ ] Yes
[x] No
```


noronhaa:
> Not 100% understanding the reason for the change. Need some explanations:
> 
> 1. Is the key loaded in all envs
> 2. Has the old one been removed
> 3. Can we change the ENV variable to more clearly align with the name of the var in the vault
> 4. If we need to change this could we rather relook at the names in general so that they have clearer prefix

The data-store-api and management-web projects' build have been failing as they use the same User which has different workbasket defaults depending which projects definition file has been imported last. These change in workbasket defaults breaks the other projects functional tests & therefor build. I have added a new user for specifically and only management-web to use to avoid these clashes. I have added this new user to vault and this PR is to get the functional tests to use this new user.

1. Key is loaded for Prod and AAT
2. Old key has not been removed as it is for an existing user being used for the data-store-api still
3. from management-web point of view the user would be assumed to be a front end user? but if it's more convention then it could be changed
4. It looks fairly clear to me, a ccd caseworker for Automation (Front End) :) However if you have suggestions though happy to look at changing it

PR #558, opened 2019-02-27
RDM-4109 Update ui toolkit version.

andrewfolga:
https://tools.hmcts.net/jira/browse/RDM-4109

### Change description ###
Update ui toolkit to pick up new changes for Case History Tab as flexible control.

```
[ ] Yes
[x] No
```


package.json
```diff
@@ -56,7 +56,7 @@
     "@angular/platform-server": "~7.0.3",
     "@angular/router": "~7.0.3",
     "@angular/upgrade": "~7.0.3",
-    "@hmcts/ccd-case-ui-toolkit": "2.27.0",
+    "@hmcts/ccd-case-ui-toolkit": "2.28.0",
```
gbenadikar: Shouldn't the bumping up of versions on case-ui-toolkit and case management be owned by QAs? They can collaborate within themselves before merging to bump up the versions

andrewfolga: Yes it should ideally be done by QAs. I just made those few steps for them :)

PR #590, opened 2019-03-12
RDM-876-1: Hide Print button when not configured

MKaramath:
https://tools.hmcts.net/jira/browse/RDM-876

### Change description ###
As a court staff I do not want to see the Print button if the print callback url is not configured so that I do not have any confusion.

ccd-case-ui-toolkit PR (related):
https://github.com/hmcts/ccd-case-ui-toolkit/pull/270

ccd-data-store-api PR (related):
https://github.com/hmcts/ccd-data-store-api/pull/401

```
[ ] Yes
[x] No
```


src/app/search/search.component.spec.ts
```diff
@@ -20,7 +20,8 @@ const CASE_TYPES: CaseType[] = [
         states: [],
         events: [],
         case_fields: [],
-        jurisdiction: null
+        jurisdiction: null,
+      printableDocumentUrl: 'print-url'
```
MSancaktutar: If this refers to a set of documents and a single url, the name should be printableDocumentsUrl.
If this refers to a set of document urls, the name should be printableDocumentUrls.
if this refers to just one document url, please ignore this comment.

MSancaktutar: This comment kindly applies to the other occurrences in the other files under this PR.

MKaramath: Good spot!   I renamed the field to printableDocumentsUrl as per the field declared in definitions API in all places.

test/features/caseView.feature
```diff
@@ -27,5 +25,14 @@ Feature: Set of scenarios testing case view functionality
       | Text Field 3        |
       | Text Field 4        |
 
+  @functional
   Scenario: print button visible on case view
-   Then the print button will be visible
+    When a case type containing every field type exists
```
noronhaa: Lets make this line more readable from a BDD point of view `Given a case type with the print button configured exist` you can then use(copy) the same implementation from the step 'a case type containing every field type exists' you will find the code in definitionFileSteps.js or better still put it in a new function and call it in both steps

test/pageObjects/caseDetailsPage.js
```diff
@@ -64,8 +64,12 @@ class CaseDetailsPage extends BasePage {
    */
 
   async isPrintButtonReady() {
```
noronhaa: can we check if this change is needed, does the original implementation of this function not return false if the button is not here?

test/features/caseView.feature
```diff
@@ -27,5 +25,14 @@ Feature: Set of scenarios testing case view functionality
...
+    And I create the case
+    Then the print button will be visible
+
+  @functional
+  Scenario: print button not visible on case view
+    When a case with Case Progression functionality exists
```
noronhaa: Similar as abive comment `Given a case with the print button not configured exists` would be better here

test/features/caseView.feature
```diff
@@ -27,5 +25,14 @@ Feature: Set of scenarios testing case view functionality
...
 
+  @functional
   Scenario: print button visible on case view
-   Then the print button will be visible
+    When a case type containing every field type exists
+    And I create the case
+    Then the print button will be visible
+
+  @functional
+  Scenario: print button not visible on case view
```
noronhaa: `print button not visible on case view by default` is more descriptive title

test/features/caseView.feature
```diff
@@ -27,5 +25,14 @@ Feature: Set of scenarios testing case view functionality
       | Text Field 3        |
       | Text Field 4        |
 
+  @functional
   Scenario: print button visible on case view
```
noronhaa: `print button visible on case view when configured` would be more descriptive title please

test/features/caseView.feature
```diff
@@ -3,8 +3,6 @@ Feature: Set of scenarios testing case view functionality
 
   Background:
     Given I have logged in
-    And a case containing Tabs functionality exists
-    And I create the case
 
   Scenario: case reference visible on case view
     Then the case reference will be visible and formatted well
```
noronhaa: don't forget to add the 2 deleted lines from the backround here - otherwise I will get confused when I come round to fixing this test :)

test/pageObjects/caseDetailsPage.js
MKaramath: Yes, we need this change.  Otherwise, it errors since the print button location do not exists

test/features/caseView.feature
MKaramath: Done.

test/features/caseView.feature
MKaramath: Done.

test/features/caseView.feature
MKaramath: Done.

test/features/caseView.feature
MKaramath: Done.

test/features/caseView.feature
MKaramath: Done.

test/pageObjects/caseDetailsPage.js
noronhaa: ok thats fine

MKaramath:
New PR created for this implementation:
https://tools.hmcts.net/jira/browse/RDM-876

PR #597, opened 2019-03-19
change-vault-key-for-fe-automated-checks

agilks:
### Change description ###

Change value key used by automated checks so the tests use a different value for user login.


ngul:
This is not needed anymore

PR #607, opened 2019-03-26
RDM-3782: Previous Case Reference data is displayed until new data is loaded

MKaramath:
https://tools.hmcts.net/jira/browse/RDM-3782

### Change description ###
Previous Case Reference data is displayed until new data is loaded

```
[ ] Yes
[x] No
```


src/app/shared/search/search-result.html
```diff
@@ -29,13 +29,13 @@ <h2 class="heading-h2">{{ caseState ? 'Case List' : 'Search result' }}</h2>
 
       <a *ngIf="colIndex == 0" routerLink="/case/{{jurisdiction.id}}/{{caseType.id}}/{{result.case_id}}"
          attr.aria-label="go to case with Case reference:{{ result.case_id | ccdCaseReference }}">
-        <ng-container class="text-16" [style.visibility]="hideRows ? 'hidden' : 'visible'">
+        <div class="text-16" [style.visibility]="hideRows ? 'hidden' : 'visible'">
```
mario-paniccia: interesting fix. I'm curious to know how changing to a div fixes the issue. Can you explain please Karamath?

MKaramath: I guess, rending ng-container element is bit heavy in comparison to div element rending, not really sure  :)

mario-paniccia: the `ng-container` never gets rendered by Angular. So the following is ignored I think:  [style.visibility]="hideRows ? 'hidden' : 'visible'"  and we always show the Reference. I wonder whether rather than replacing it with a div, which could cause some side effects, we could replace `[style.visibility]` with `ngIf`. Can you give it a try please?

PR #634, opened 2019-04-10
RDM-2348-2: Grey bar styling

MKaramath:
https://tools.hmcts.net/jira/browse/RDM-2348

### Change description ###
Grey bar show and hide styling

```
[ ] Yes
[x] No
```


test/features/conditionalLogic.feature
```diff
@@ -12,7 +12,10 @@ Feature: Set of scenarios testing show & hide functionality and conditional logi
   Scenario: field visible when show & hide condition is met
     Given a case type containing conditionals exists
     When I enter 'showmethemoney' into the 'text' field
-    Then the field with label 'Text Field 2' is visible
+    Then the field with label 'Text Field 2' is visible with grey bar
+    And I click the Continue button
+    And the field with label 'Text Field 3' is visible
+    And the field with label 'Text Field 3' is visible without grey bar
```
andrewfolga: Why two similarly looking steps? Wouldn't it be enough to have the last one only?

veeresha-hmcts: @andrewfolga I think that was done because we wanted to have two asserts, one for the text field visible and there other for field not having a grey bar. We could combine those two but I thought keeping it separate will probably help min reusing the methods somewhere else. We will review this again later.

test/pageObjects/createCaseWizardPage.js
```diff
@@ -13,6 +13,7 @@ class CreateCaseWizardPage extends BasePage{
       this.answerValueXpathTemplate = '//span[text()="LABEL-TEXT-PLACEHOLDER"]/../following-sibling::td//ccd-field-read-label/*';
       this.answerChangeLinkXpathTemplate = '//span[text()="LABEL-TEXT-PLACEHOLDER"]/../../td[2]/a';
       this.fieldLabels = 'fieldset span';
+      this.greyBarfieldLabels = '.show-condition-grey-bar span';
```
andrewfolga: Please spell out variable names camel case.

test/features/conditionalLogic.feature
andrewfolga: In that case should not the last step sound more like:
`And the field with label 'Text Field 3' is without grey bar`
just omitting the visible?

andrewfolga: Also it is breaking the not more then 5 steps per scenario rule.

andrewfolga:
Added more comments

test/features/conditionalLogic.feature
MKaramath: Done.

test/pageObjects/createCaseWizardPage.js
MKaramath: Done.

test/features/caseList.feature
```diff
@@ -1,4 +1,4 @@
-@functional @caseList
+@functional @caseList @broken
```
andrewfolga: why marking this?

test/features/fieldTypes.feature
```diff
@@ -97,7 +97,7 @@ Feature: Set of scenarios to check we can read and write to all field data types
     Then there will be validation errors
     And the 'Continue' button will be disabled
 
-    @validation @regex
+    @validation @regex @broken
```
andrewfolga: why marking this?

PR #637, opened 2019-04-11
RDM-4184 Update func tests and spreadsheet to reflect removing of the…

andrewfolga:
… fixed History tab.

Please remove this line and everything above and fill the following sections:

https://tools.hmcts.net/jira/browse/RDM-4184

### Change description ###
Update func tests and spreadsheet to reflect removing of Fixed History Tab

```
[ ] Yes
[x] No
```


noronhaa:
What will the mean for our definition file, will we need to update the definition file to have a tab called 'history' on the case type otherwise the tests will fail? if so has this been done already?

PR #645, opened 2019-04-15
RDM-3460 Workbasket input search - use elementPath to build caseField id

rafalkalita:
https://tools.hmcts.net/jira/browse/RDM-3460

### Change description ###
Complex element *only* as list screen input

```
[ ] Yes
[X ] No
```


src/app/workbasket/filters/workbasket-filters.component.ts
```diff
@@ -145,15 +145,16 @@ export class WorkbasketFiltersComponent implements OnInit {
             .sort(this.orderService.sortAsc);
           const formValue = this.windowService.getLocalStorage(FORM_GROUP_VAL_LOC_STORAGE);
 
-          for (let i = 0; i < workbasketInputs.length; i++) {
-            let item = workbasketInputs[i];
+          workbasketInputs.forEach(item => {
+            if (item.field.elementPath) {
+              item.field.id = item.field.id + '.' + item.field.elementPath;
```
mario-paniccia: can't see elementPath added to the Field model in this PR

mario-paniccia: changing the properties of a model class this way is not the safest thing to do. Especially if that property is the `id`. In this case I don't think it's an issue, because the WorkbasketInput has its own 'copy' of a Field. But still it could introduce nasty side effects. If we want to play it safe, an alternative would be to add a new property on the WorkbasketInput called something like `IdForSearch`, which by default is equal to the `id`. Then when we send the search requests, we use the `IdForSearch` rather than the `id`

mario-paniccia: furthermore, do we need to add `elementPath` to Field? it's a quite specific property needed only for this particular use case. Can't we have it on the workbaskeInput model itself instead?

rafalkalita: Field model is coming from the ui-toolkit. It will appear when ui-toolkit will get merged.

PR #658, opened 2019-04-23
Rdm 4535 new contract

adnanakgun:

https://tools.hmcts.net/jira/browse/RDM-4535
https://tools.hmcts.net/jira/browse/RDM-4576

### Change description ###

Added router actions to the consumer component and removed DefinitionService

```
[ ] Yes
[x] No
```


src/app/workbasket/workbasket.component.html
```diff
@@ -3,9 +3,9 @@
 
     <div class="display-left" leftBody>
       <ccd-workbasket-filters
```
andrewfolga: Seems like after this is merge we will be using the new contract in man web. This seems wrong to me as we will be making a call for data we already have in `profile.jurisdictions`.

src/app/workbasket/workbasket.component.html
```diff
@@ -3,9 +3,9 @@
...
-        [jurisdictions]="profile.jurisdictions"
         [defaults]="profile.default.workbasket"
         (onApply)="applyFilter($event)"
+        (onReset)="applyReset($event)"
```
andrewfolga: 👍 for moving the navigation out of the extracted component into consumer.

src/app/workbasket/workbasket.component.html
adnanakgun: This has now been regressed.

PR #663, opened 2019-04-30
[PREVIEW] Rdm 4563 ai for web pages

MSancaktutar:
Jira link for this change:
https://tools.hmcts.net/jira/browse/RDM-4563

```
[ ] Yes
[X] No
```


server.ts
```diff
@@ -42,9 +42,15 @@ const CONFIG = {
   'activity_next_poll_request_ms': parseInt(process.env['CCD_ACTIVITY_NEXT_POLL_REQUEST_MS'], 10) || 5000,
   'activity_retry': parseInt(process.env['CCD_ACTIVITY_RETRY'], 10) || 5,
   'activity_batch_collection_delay_ms': parseInt(process.env['CCD_ACTIVITY_BATCH_COLLECTION_DELAY_MS'], 10) || 1,
-  'activity_max_request_per_batch': parseInt(process.env['CCD_ACTIVITY_MAX_REQUEST_PER_BATCH'], 10) || 25
+  'activity_max_request_per_batch': parseInt(process.env['CCD_ACTIVITY_MAX_REQUEST_PER_BATCH'], 10) || 25,
+  'web_pages_instrumentation_key': process.env['WEB_PAGES_APPINSIGHTS_INSTRUMENTATION_KEY'] || '',
```
timja: Please don't use another environment variable name, we already have 3,
https://github.com/hmcts/cnp-module-webapp/blob/master/main.tf#L64-L69

infrastructure/main.tf
```diff
@@ -55,6 +55,7 @@ module "case-management-web" {
   website_local_cache_sizeinmb = 500
   capacity = "${var.capacity}"
   appinsights_instrumentation_key = "${var.appinsights_instrumentation_key}"
+  web_pages_appinsights_instrumentation_key = "${var.web_pages_appinsights_instrumentation_key}"
```
timja: this isn't a thing, the above key will be used

server.ts
MSancaktutar: The other 3 are quite similar to each other. With this one, I wanted to have an express separation between the key for the AI resource of the BE service and the key for that of the FE pages. These keys may be different indeed (as suggested by MS as well).

infrastructure/main.tf
MSancaktutar: I wanted to have an express separation between the key for the AI resource of the BE service and the key for that of the FE pages. These keys may be different indeed (as suggested by MS as well).

timja: pass it through as an app setting then not as a module parameter

server.ts
```diff
@@ -42,9 +42,15 @@ const CONFIG = {
...
 };
 
+const fs = require('fs');
+const key_file_name = './browser/ai-instrumentation-key.js';
```
mario-paniccia: camel case maybe?

src/public/ai-client-setup.js
```diff
@@ -0,0 +1,44 @@
+var appInsights = window.appInsights
+		|| function(a) {
```
mario-paniccia: has this been copied from somewhere? 
this code is quite cryptic

infrastructure/variables.tf
```diff
@@ -87,6 +87,10 @@ variable "common_tags" {
   type = "map"
 }
 
+variable "web_pages_appinsights_instrumentation_key" {
```
mario-paniccia: can we have this enabled on all environments but leave it disabled for now on Prod?

infrastructure/main.tf
dwaynebailey: MS also suggests one AppInsights across all apps.  I'd prefer 1 tbh as its a single place for all info.  Since this is new lets just use the one we have, we can always separate it out if it proves to not work for us.  As it stands this code won't actually tie into any AppInsights instance that we have.

infrastructure/sprod.tfvars
```diff
@@ -6,4 +6,3 @@ asp_name = "ccd-case-management-web-sprod"
 asp_rg = "ccd-case-management-web-sprod"
 
 idam_authentication_web_url = "http://idam-web-public-idam-sprod.service.core-compute-idam-sprod.internal"
-
```
dwaynebailey: Please no noop changes

infrastructure/demo.tfvars
```diff
@@ -4,4 +4,4 @@ document_management_url = "^https?://(?:api-gateway\\.preprod\\.dm\\.reform\\.hm
 ccd_print_service_url = "https://return-case-doc.ccd.demo.platform.hmcts.net"
 activity_enabled = "true"
 external_host_name = "www-ccd.demo.platform.hmcts.net"
-capacity = "2"
+capacity = "2"
```
dwaynebailey: Please no noop changes

infrastructure/main.tf
MSancaktutar: Dwayne, our 1 App Insights resource is collecting everything from Azure-based applications. If we use the inst key of that one for Web Page AI integration, then the known and accepted vulnerabilities will have been accepted to impact the AI for all others. For details, please see:
https://tools.hmcts.net/confluence/display/RCCD/Application+Insights+for+Angular+Clients+on+Browsers
and
https://docs.microsoft.com/en-us/azure/azure-monitor/app/troubleshoot-faq#my-instrumentation-key-is-visible-in-my-web-page-source

On the second link, as you'll see, using a separate AI resource dedicated to web pages is also a mentioned in MS's Q&A page. If we expose the inst key of all-in-one AI resource to public, that may make security approval of this change a lot more difficult.

src/public/ai-client-setup.js
MSancaktutar: The code is from Microsoft, here: https://docs.microsoft.com/en-us/azure/azure-monitor/app/sharepoint

It is kind of an obfuscated code and was much uglier before I formatted it on IDE. :)

server.ts
MSancaktutar: made it uppercase, rather.

infrastructure/variables.tf
MSancaktutar: sure.

infrastructure/sprod.tfvars
MSancaktutar: undone

infrastructure/demo.tfvars
MSancaktutar: undone

infrastructure/main.tf
timja: They don't say that you should do it though.
```
This is common practice in monitoring solutions.
It can't be used to steal your data.
It could be used to skew your data or trigger alerts.
We have not heard that any customer has had such problems.
```

I think this is inventing a solution when we don't need it?

PR #677, opened 2019-05-16
RDM-4622-Functional Test 

ganganna-peddapothu:
**Before creating a pull request make sure that:**

### Page Show Conditions not working as expected in CaseEventToFields - functional tests ###

```
[ ] Yes
[ X] No
```


test/stepDefinitions/caseDetailsSteps.js
```diff
@@ -97,9 +97,12 @@ defineSupportCode(function ({ Given, When, Then, Before, After }) {
   });
 
   When(/^I click on its first accordion on the '(.*)' tab$/, async function (tabName) {
-    await caseDetailsPage.clickTab(tabName);
-    await caseDetailsPage.clickFirstAccordian();
+      await caseDetailsPage.clickTab(tabName);
+      await caseDetailsPage.clickFirstAccordian();
   });
 
+  When(/^I progress to the next event which contains read-only value of the selected fixed list value$/, async function() {
```
mario-paniccia: there is a mismatch between that this method is doing and the step description. For example there's nothing regarding read only or fixed lists in the method implementation.
We should make the step description more generic. If we do that we would also improve reusability

test/stepDefinitions/create-caseSteps.js
```diff
@@ -229,4 +229,28 @@ defineSupportCode(function ({ Given, When, Then}) {
...
+    let fieldPresent = await caseWizardPage.isFieldPresent('fixed-list')
+    expect(fieldPresent).to.be.false;
+  });
+
+  When(/^I continue to the 3rd page$/, async function() {
+    await caseWizardPage.clickContinueButton();
+    await caseWizardPage.clickContinueButton();
+  });
+
+  Then(/^The page that satisfies show condition of the fixed list item value is displayed$/, async function() {
```
mario-paniccia: there is a mismatch between what this method is doing and its step description. What the implementation does is checking the page header matches an expected value. 
We should make the step description more generic and aligned to the implementation. If we do that then we would also improve reusability
Also to improve the step reusability, we should avoid hardcoding values ('Conditional Page 3 - Approve a case')

test/stepDefinitions/caseDetailsSteps.js
```diff
@@ -97,9 +97,12 @@ defineSupportCode(function ({ Given, When, Then, Before, After }) {
   });
 
   When(/^I click on its first accordion on the '(.*)' tab$/, async function (tabName) {
-    await caseDetailsPage.clickTab(tabName);
-    await caseDetailsPage.clickFirstAccordian();
+      await caseDetailsPage.clickTab(tabName);
+      await caseDetailsPage.clickFirstAccordian();
   });
 
+  When(/^I progress to the next event '(.*)'$/, async function(event) {
```
mario-paniccia: looking at the code, should't this be 'I start the event' or 'I trigger the event'?

test/stepDefinitions/create-caseSteps.js
```diff
@@ -179,10 +179,18 @@ defineSupportCode(function ({ Given, When, Then}) {
   });
 
   Then(/^I will be on the '(.*)' page$/, async function (expectedPageHeader) {
-    let pageHeader = await caseWizardPage.getPageHeader();
-    expect(pageHeader).to.equal(expectedPageHeader);
+    await IAmOnPageWithHeader(expectedPageHeader)
+  })
+
+  Then(/^the '(.*)' page should be displayed as a result of the fixed list show condition$/, async function(expectedPageHeader) {
```
mario-paniccia: looking at the code, this should be: ```/^the '(.*)' page should be displayed``` only. The part ``` as a result of the fixed list show condition``` limits the reusability of this step. As a result we will have to introduce many duplicated versions in the future

test/stepDefinitions/create-caseSteps.js
```diff
@@ -229,4 +237,23 @@ defineSupportCode(function ({ Given, When, Then}) {
...
+    await caseWizardPage.clickContinueButton();
+    await caseWizardPage.clickSubmitCaseButton();
+  });
+
+  When(/^The fixed list item is hidden on the first page$/, async function() {
+    let fieldPresent = await caseWizardPage.isFieldPresent('fixed-list')
+    expect(fieldPresent).to.be.false;
+  });
+
+  When(/^I continue to the 3rd page$/, async function() {
```
mario-paniccia: mismatch between step description and implementation. We either change the implementation to jump to the page number indicated in the description, or we change the description to be: `I move forward 2 pages`

test/stepDefinitions/emptySteps.js
```diff
@@ -0,0 +1,15 @@
+
+var { defineSupportCode } = require("cucumber");
+
+/**
+ * This class is to put steps that require no implementation but are there purely for readability on the BDD side
+ */
+defineSupportCode(function ({ Given, When, Then}) {
+
+  Given(/^this event contains a read only value of the fixed list from the previous event$/, function() {
+    //The (Approve case) event has been configured in the definition file to have a read only field containing the value
```
mario-paniccia: is this a good practice?

test/features/conditionalLogic.feature
```diff
@@ -58,4 +58,12 @@ Feature: Set of scenarios testing show & hide functionality and conditional logi
     When I do NOT meet the condition for showing the field in the tab
     Then the 'Text Field 3' field will NOT be visible on the 'Conditional Tab 1' tab
 
-
+  @bugfix @RDM-4622
+  Scenario: Page Show Condition works for read only field in previous event
+    Given a case type containing conditionals exists
```
mario-paniccia: not sure if it's me, but I don't find this scenario clear enough. I've asked Aman and he can't understand it as well. I think we should improve it

test/stepDefinitions/caseDetailsSteps.js
noronhaa: this is just being picky but I will give it to you

test/stepDefinitions/create-caseSteps.js
noronhaa: sure- but there comes a point where you need to strike the balance between giving details of what is happening in a step in the context of the whole scenario over re-usability, abstracting the main code into a function means you are not having to write duplicated code while being able to express more details in the bdd

test/stepDefinitions/emptySteps.js
noronhaa: I have taken this out for now however if you generalise a scenario to much it can be good for reusability but can sometimes loose the more niche edge case details of whats going on, this detail was originally abstracted out of another step to make it more reusable but this line is the defining point of the test, or the what caused the bug in the first place depending on how you want to look at it. 

I would say it's not bad practise to do this if it is bringing value to the reader but there may not necessarily be anything to implement, since we should typically be coming up with bdds in 3As there may be situations where there is no relevant code implementation to a step (perhaps it's been done a definition config already for example) but having the blank step there makes the scenario read better

test/features/conditionalLogic.feature
noronhaa: have broken it into 2 smaller scenarios

PR #684, opened 2019-05-23
Rdm-4535 Case List Filters new contract

noronhaa:
### Change description ###
Move Case List filters to toolkit and implements new contract 

```
[ ] Yes
[x] No
```


src/app/workbasket/filters/workbasket-filters.component.spec.ts
```diff
@@ -1,1058 +0,0 @@
...
-  let router: any;
-  let activatedRoute: any;
-  let jurisdictionService: JurisdictionService;
-  let workbasketInputFilterService: any;
-  let orderService: any;
-  let alertService: AlertService;
-  let windowService;
-  const TEST_FORM_GROUP = new FormGroup({});
-
-  describe('Clear localStorage for workbasket filters', () => {
```
andrewfolga: These tests are removed here and not added in the ui toolkit for case-list-filters.components.ts. Are we getting rid of them?

src/app/workbasket/filters/workbasket-filters.component.spec.ts
```diff
@@ -1,1058 +0,0 @@
...
-      component.selected.jurisdiction = JURISDICTION_2;
-      component.selected.caseType = null;
-      component.selected.caseState = null;
-
-      fixture.detectChanges();
-      let button = de.query($APPLY_BUTTON);
-      expect(button.nativeElement.disabled).toBeTruthy();
-    }));
-
-    it('should have an Apply button disabled when case type is not set', async(() => {
```
andrewfolga: Some other tests like this one assessing the buttons being disabled are missing in `case-list-filters.component.spec.ts`. Are we sure we do not want to keep them?

src/app/workbasket/workbasket.component.ts
```diff
@@ -94,6 +98,14 @@ export class WorkbasketComponent implements OnInit {
         });
 
     this.scrollToTop();
+
+    this.router.navigate(['/list/case'], {
```
andrewfolga: Do we have a test case asserting this action?

src/app/workbasket/workbasket.component.ts
```diff
@@ -94,6 +98,14 @@ export class WorkbasketComponent implements OnInit {
...
+      queryParams: returnValue.queryParams
+    });
+  }
+
+  applyReset() {
+    this.router.navigate(['/list/case']);
```
andrewfolga: Can we have a test case asserting this action?

src/app/workbasket/workbasket.component.ts
adnanakgun: Added now

src/app/workbasket/workbasket.component.ts
adnanakgun: Added now

PR #696, opened 2019-06-03
RDM-4877 Functional tests for show and hide for basic types

rafalkalita:
https://tools.hmcts.net/jira/browse/RDM-4877

### Change description ###
This is a bunch of tests to cover show and hide functionality for different basic field types.

```
[ ] Yes
[ X] No
```


test/utils/fieldUtils.js
```diff
@@ -52,14 +53,29 @@ class FieldDataTypes {
...
   }
 
+  /**
+   * Sets the value into the AddressLine1 field of the address field specified by id.
+   * If no id is specified it will populate AddressLine1 of the first found address.
+   * @param value
+   * @param id
+   * @returns CCDAddressUKField Object
+   */
+  async enterIntoAddressLine1Field(value, id) {
```
noronhaa: This works for now but I wonder in the future if this should be refactored to just 'enter into addressUK field' and should pass an object or an array which would then be resolved in the CCDAddressUKField class and map tot he fields in the complexType

test/stepDefinitions/validationSteps.js
```diff
@@ -66,12 +66,16 @@ defineSupportCode(function ({ Given, When, Then, Before, After }) {
...
+    await baseSteps.navigateToCreateCasePage();
+    await caseWizardPage.interactWithField(fieldType, value);
   });
 
   When(/^I re-enter '(.*)' into the '(.*)' field$/, async function (value, fieldType) {
-    await caseWizardPage.interactWithField(fieldType,value);
+    await caseWizardPage.interactWithField(fieldType, value);
+  });
+
+  When(/^I re-enter '(.*)' into the '(.*)' field with id '(.*)'$/, async function (value, fieldType, id) {
```
noronhaa: This looks like a new Step but I don't see it used in the BDD so is it needed any longer or it can be removed? (Also we should try to not to parse IDs in form the BDD but have them coming from the page object level instead)

rafalkalita: Yes. Will remove it. I was just testing if caseWizardPage.interactWithField(fieldType, value, id) will work.

test/utils/fieldUtils.js
rafalkalita: Yes, I have left the rest of the fields for later.

PR #700, opened 2019-06-04
Rdm-4789 Pagination functional tests

noronhaa:
https://tools.hmcts.net/jira/browse/RDM-4789

### Change description ###
Adding Func Tests for Pagination component on Case List filters

```
[ ] Yes
[X] No
```


test/features/caseList.feature
```diff
@@ -8,3 +8,72 @@ Feature: Set of Scenarios testing the functionality of the Case List table
     Given a case type exists with case reference configured in the case list results
     And there are cases listed on the case list page for that case type
     Then the case reference is displayed in the case list results with hyphens
+
+  @pagination
+  Scenario: No Previous button on pagination when on page 1
```
mario-paniccia: considering that ui func tests are  a bit slow, would it make sense to combine some of these scenarios into 1 or 2?
For example 1 scenario where you start on page 1 and verify all the stuff on page 1 then in the same scenario move to other pages and do the various verifications etc..

noronhaa: Appreciate your concern about speed of tests however it's not good practice to bundle many things into a single test, if the first 'thing' you are testing fails you will have no knowledge  that the next few things in the test are broken or not as the test has failed early. 

These tests take 1min 38secs to run when I ran against AAT and this feature file would also be running in parallel with other features so not a huge hit on execution time. However, I understand the bigger picture that speed of tests is important and that we will get to a point where they will start to take too long but rather than changing the format of out tests there are other things we can do to help with execution time (eg run more in parallel, use cookie injection to bypass login form) which we can look into when needed

mario-paniccia: rules for unit tests don't always apply to UI func tests. UI func tests usually are more coarse grained than unit tests. You can easily bundle all those into 1 test, if it fails you can understand where it fails from the step in error. 
Consider that in a previous project we used to write 1 scenario that was testing a multi page functionality and that means it had to navigate between various pages and verify multiple stuff. If we breaked that into single little tests it'll be much less understandable and valuable

test/features/caseList.feature
```diff
@@ -8,3 +8,45 @@ Feature: Set of Scenarios testing the functionality of the Case List table
     Given a case type exists with case reference configured in the case list results
     And there are cases listed on the case list page for that case type
     Then the case reference is displayed in the case list results with hyphens
+
+  @pagination
+  Scenario: Next and previous buttons appear appropriatley based on pagination position selected
+    Given I am on case list page
+    And there are more than 1 page of results
+    And I am on page 1 of results
+    And the 'Previous' button will not be displayed on pagination
```
mario-paniccia: Then?

test/features/caseList.feature
```diff
@@ -8,3 +8,45 @@ Feature: Set of Scenarios testing the functionality of the Case List table
...
+    And I navigate to page 2 of results
```
mario-paniccia: Given I navigate...?

test/features/caseList.feature
```diff
@@ -8,3 +8,45 @@ Feature: Set of Scenarios testing the functionality of the Case List table
...
+    And the 'Previous' button will be displayed on pagination
```
mario-paniccia: Then the 'Previous'....?

test/features/caseList.feature
```diff
@@ -8,3 +8,45 @@ Feature: Set of Scenarios testing the functionality of the Case List table
...
+    When I am on the last page of results
+    Then the 'Next' button will not be displayed on pagination
+
+  @pagination
+  Scenario: Clicking Next/Previous button will navigate to respective next/previous page and show that page selected on pagination
+    Given I am on case list page
+    And there are more than 1 page of results
+    And I am on page 1 of results
+    And I click the pagination 'Next' button
+    And I will see a different page of results
```
mario-paniccia: Then?

test/features/caseList.feature
```diff
@@ -8,3 +8,45 @@ Feature: Set of Scenarios testing the functionality of the Case List table
...
+    And page '2' will be selected on the pagination
```
mario-paniccia: When?

test/features/caseList.feature
```diff
@@ -8,3 +8,45 @@ Feature: Set of Scenarios testing the functionality of the Case List table
...
+    And I click the pagination 'Previous' button
+    And I will see a different page of results
```
mario-paniccia: Then?

test/features/caseList.feature
```diff
@@ -8,3 +8,45 @@ Feature: Set of Scenarios testing the functionality of the Case List table
...
+    Then page '1' will be selected on the pagination
```
mario-paniccia: And?

test/features/caseList.feature
```diff
@@ -8,3 +8,45 @@ Feature: Set of Scenarios testing the functionality of the Case List table
...
+
+  @pagination
+  Scenario: Clicking 2 button on pagination will navigate to 2nd page and show 2 page selected on pagination
+    Given I am on case list page
+    And there are more than 1 page of results
+    When I click the pagination '2' button
```
mario-paniccia: I'd merge this with the above. They are both testing navigation functionality.

PR #702, opened 2019-06-10
RDM-4833 Showing fields within the same page

rafalkalita:
https://tools.hmcts.net/jira/browse/RDM-4833

### Change description ###
A set of tests for show and hide using same page data to show or hide the fields.
- from a simple field type
- using field in a complex
- using field in a collection
- with a mix complex of collection of complex

```
[ ] Yes
[ X] No
```


test/features/complexTypes.feature
```diff
@@ -9,16 +9,6 @@ Feature: Set of scenarios testing show & hide functionality and conditional logi
     When I populate the form with the school data
     Then only the fields defined in EventToComplexTypes sheet should be visible
 
-  Scenario: hides complex element nested in collections when showCondition uses top element data
```
rafalkalita: Moved both to complexTypes.feature

test/features/conditionalLogic.feature
```diff
@@ -4,16 +4,6 @@ Feature: Set of scenarios testing show & hide functionality and conditional logi
   Background:
     Given I have logged in
 
-  Scenario: field NOT visible when show & hide condition is NOT met
```
rafalkalita: Covered by basicFields.feature

test/features/showandhide/basicFields.feature
```diff
@@ -0,0 +1,39 @@
+@functional @showandhide
+Feature: Hide and show of basic fields
+
+  Background:
+    Given I have logged in
+
+  Scenario Outline: fields on the same page visible when show & hide condition is met based on mandatory and optional fields
+    Given a case type containing conditionals exists
+    And I start createCase event
+    When I enter '<value>' into the text field with id '<fieldId>'
```
mario-paniccia: should we have  a step that verifies the fields are initially hidden?

test/features/showandhide/basicFields.feature
```diff
@@ -0,0 +1,39 @@
...
+    Then the field with label '<expectedLabel>' is visible
+
+    Examples:
```
mario-paniccia: was wondering whether we could add a 4th column where we display the show condition. It would help better understanding the test. Even thought is a bit of maintainability problem because we have to remember to update
that column if we change the show condition in the spreadsheet. 
Worth thinking about it

test/features/showandhide/basicFields.feature
```diff
@@ -0,0 +1,39 @@
...
+      | value            | fieldId           | expectedLabel          |
+      | showmethemoney   | TextField         | Text Field 2           |
+      | showme           | TextFieldOptional | Text Field ShowHide 13 |
+
+  Scenario Outline: fields on the same page NOT visible when show & hide condition is NOT met based on mandatory and optional fields
+    Given a case type containing conditionals exists
+    And I start createCase event
```
mario-paniccia: we should verify the fields are initially visible

test/features/showandhide/basicFields.feature
```diff
@@ -0,0 +1,39 @@
...
+
+  Scenario Outline: fields on the same page visible when show & hide condition is met based on mandatory and optional fields
+    Given a case type containing conditionals exists
+    And I start createCase event
+    When I enter '<value>' into the text field with id '<fieldId>'
+    Then the field with label '<expectedLabel>' is visible
+
+    Examples:
+      | value            | fieldId           | expectedLabel          |
+      | showmethemoney   | TextField         | Text Field 2           |
```
mario-paniccia: if we introduce the step I suggested above, we could reuse that same step here to do the verification. i.e. 
```
Then I expect the page to be in the following state:
| fieldId | value |
```

test/features/showandhide/basicFields.feature
rafalkalita: That is why it is so time consuming to analyze the tests. We could have a comment in the .feature file saying what is in the spreadsheet for a given test. It would be better than adding an additional step.
I have had a chat with few people already. I think the solution is to generate Excel definition from the builder in the code. But this is a bigger task and we would need our PRs to host the back-end on AKS.

rafalkalita: This is not trivial to implement. You can easily say if the field is hidden or not by id, but the information about showCondition would have to be accessed from the Angular model.

mario-paniccia: it's not trivial but it would be worth the investment of time in my opinion. You'll have a highly reusable step that can be used instead of many other wordy steps. That helps document the test. And it's always up to date, something a comment can't guarantee. 
I don't see why it's difficult in the step to verify 
`| visibility | show condition | initially hidden | `
Can't you grab the case type definition? easy easy

rafalkalita: I am not sure we are on the same page. The definition is on AAT, and the test is run in AKS and only has access to what is exposed on the UI. I cannot see how we could access `show condition`. Visibility we check by getting all labels and testing if the given label is in the array. But the show condition is in the Angular model and we execute JS agains it. If you have any ideas please share.

test/features/showandhide/basicFields.feature
rafalkalita: Done

test/features/showandhide/basicFields.feature
rafalkalita: In this case the field is initially invisible. It is because the showCondition is set to hide it when its value is not set to 'showmethemoney'.

test/stepDefinitions/create-caseSteps.js
```diff
@@ -209,18 +211,18 @@ defineSupportCode(function ({ Given, When, Then}) {
       });
   });
 
-  async function populateFormDataWithSupportFieldSetTo(supportAnswer) {
+  async function populateFormDataWithSupportFieldSetTo(supportAnswer = 'Yes', className = 'A team', isClassMemeberAutistic = 'Yes') {
```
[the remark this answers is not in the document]
rafalkalita: Functions default parameters - some JS magic :)

test/features/showandhide/basicFields.feature
mario-paniccia: not sure why this is so difficult. When we launch the suite of func tests we retrieve the case type definition once from Def Store via it's rest API `getCaseType`

mario-paniccia: Here's what I'm suggesting:
```
Scenario Outline: fields on the same page visible when show & hide condition is met based on mandatory and optional fields
    Given a case type containing conditionals exists
    And I start createCase event
    And the page contains the following fields
    Examples:
      | fieldId               | value             | display context    | show condition                   |  shown  |
      | TextField             |                   | Mandatory          |                                  |  yes    |
      | TextFieldOptional     |                   | Optional           |                                  |  yes    |
      | TextFieldShowHide13   |                   | Optional           | TextFieldOptional="showme"       |  no     |

    When I enter '<value>' into the text field with id '<fieldId>'

    Examples:
      | value            | fieldId           |
      | showmethemoney   | TextField         |
      | showme           | TextFieldOptional |

    Then the page contains the following fields
    Examples:
      | fieldId               | value             |  shown  |
      | TextField             | showmethemoney    |  yes    |
      | TextFieldOptional     | showme            |  yes    |
      | TextFieldShowHide13   |                   |  yes    |
```

test/pageObjects/wizardPages/conditionals_CreateCase_ConditionalPage1.js
```diff
@@ -0,0 +1,49 @@
+let CCDStringField = require('../ccd-components/fields/ccdStringField.js');
+
+class ConditionalsCreateCasePage1WizardPage extends CreateCaseWizardPage {
+
+  constructor() {
+    super();
+    this._fieldUsedInShowCondition = ''; // TextField | TextFieldOptional
+    this.mandatoryTextField = new CCDStringField('ccd-write-text-field', 'text', 'TextField');
```
noronhaa: Nothing wrong with this, but nothing stopping us just passing the ID in as the first param as either `#TextField` or `ccd-write-text-field #TextField` not sure why we have a id param when we already have a css param that an id can be parsed into. (Just some general thoughts)

test/stepDefinitions/create-caseSteps.js
```diff
@@ -14,6 +15,7 @@ var { defineSupportCode } = require("cucumber");
 defineSupportCode(function ({ Given, When, Then}) {
 
   let caseWizardPage = new CreateCaseWizardPage();
+  let createCasePage1 = new ConditionalsCreateCasePage1();
```
noronhaa: We may want to rename the var to conditionalsCreateCasePage1 to reduce confusion later on when more page object classes are being called

PR #723, opened 2019-06-26
RDM-5211 - using idam-pr chart to add redirect url for PRs

smathangi:

https://tools.hmcts.net/jira/browse/RDM-5211

### Change description ###
IDAM redirect url for PR environments
https://tools.hmcts.net/confluence/display/SISM/AKS+set+up+for+service+applications+PRs

```
[ ] Yes
[ x] No
```

### DIFF
--- a/charts/ccd-case-management-web/requirements.yaml
+++ b/charts/ccd-case-management-web/requirements.yaml
@@ -3,7 +3,7 @@ dependencies:
     version: 1.0.2
     repository: '@hmcts'
   - name: idam-pr
-    version: ~1.0.0
+    version: ~2.0.0
     repository: '@hmcts'
     tags:
     - idam-pr
--- a/charts/ccd-case-management-web/requirements.yaml
+++ b/charts/ccd-case-management-web/requirements.yaml
@@ -2,3 +2,8 @@ dependencies:
   - name: nodejs
     version: 1.0.2
     repository: '@hmcts'
+  - name: idam-pr
+    version: ~1.0.0
+    repository: '@hmcts'
+    tags:
+    - idam-pr
--- a/charts/ccd-case-management-web/requirements.yaml
+++ b/charts/ccd-case-management-web/requirements.yaml
@@ -1,6 +1,6 @@
 dependencies:
   - name: nodejs
-    version: 1.0.2
+    version: ~1.2.0
     repository: '@hmcts'
   - name: idam-pr
     version: ~2.0.0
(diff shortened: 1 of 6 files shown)


charts/ccd-case-management-web/values.template.yaml
```diff
@@ -13,3 +13,10 @@ nodejs:
     CCD_ACTIVITY_RETRY: 5
     CCD_ACTIVITY_BATCH_COLLECTION_DELAY_MS: 1
     CCD_ACTIVITY_MAX_REQUEST_PER_BATCH: 25
+
```
mario-paniccia: this should be applied to preview env only. Since we don't have a dedicated file for preview, wonder whether this will impact other envs as well

smathangi: ideally yes :). I guess we also need to change values.template.yaml to values.yaml as there is Jenkins warning atm. 
Eg: https://build.platform.hmcts.net/job/HMCTS_CDM/job/ccd-case-management-web/view/change-requests/job/PR-722/3/consoleFull
search for "values.yaml".  I'll change both of them

charts/ccd-case-management-web/values.preview.template.yaml
```diff
@@ -0,0 +1,22 @@
+tags:
```
timja: values.yaml should not include any templated values, the templated values that are currently there should be moved here and sensible default should be set there such as the latest image.

mario-paniccia: 👍

mario-paniccia: @smathangi we could address Tim's comment as part of a separate PR dedicated to that issue perhaps? To minimise the risks?

smathangi: Thanks @timja . Can you take a look again please.

charts/ccd-case-management-web/requirements.yaml
```diff
@@ -2,3 +2,8 @@ dependencies:
   - name: nodejs
     version: 1.0.2
     repository: '@hmcts'
+  - name: idam-pr
```
timja: Can you also bump this chart version up in Chart.yaml please

timja: I actually meant the ccd-case-management-web chart version, but the bump you did was a good bump. but can you update Chart.yaml too please

smathangi: Thanks. Actually i didn't read full comment. Bumped now :)

PR #710, opened 2019-06-17
Rdm 4997 FE functional tests

andrewfolga:
### JIRA link (if applicable) ###
https://tools.hmcts.net/jira/browse/RDM-4997

### Change description ###
FE functional tests

**Does this PR introduce a breaking change?** (check one with "x")

```
[ ] Yes
[x] No
```

PR #727, opened 2019-06-27
RDM 5022 - implement new cucumber step to improve test readability

rafalkalita:
https://tools.hmcts.net/jira/browse/RDM-5022

### Change description ###
Added Given steps with tabular data matching excel definition data.
skipped conditionalLogic.feature, because it needs reorganising and will be handled as a part of RDM-4820 tasks: visibility of tabs, pages, fields in tabs, fields in pages, fields in CYA etc. 

```
[ ] Yes
[ X] No
```


test/features/caseView.feature
```diff
@@ -30,15 +44,3 @@ Feature: Set of scenarios testing case view functionality
       | Text Field 2        |
       | Text Field 3        |
       | Text Field 4        |
-
-  @functional
```
noronhaa: Are we completely deleting these Scenarios for some reason? or was this a mistake?

test/features/caseView_printButton.feature
```diff
@@ -0,0 +1,23 @@
...
+Feature: Set of scenarios testing case view functionality
+
+  Background:
+    Given I have logged in
+
+  Scenario: print button visible on case view when configured
+    Given the definition sheet 'CaseType' looks like this
+      | ID              | Name                 | PrintableDocumentsUrl                                                                     |
+      | AllDataTypes2   | All Field Data Types | http://localhost:4452/callback/jurisdictions/AUTOTEST1/case-types/AllDataTypes2/documents |
+      | CaseProgression | Case Progression | |
```
noronhaa: lets make the datatable pretty and in line with the other rows here :)

test/features/search.feature
```diff
@@ -2,6 +2,36 @@
...
+      | AllDataTypes2 | MoneyField       | Money Field        | MoneyGBP          |                    |
+      | AllDataTypes2 | DocumentField    | Document Field     | Document          |                    |
+      | AllDataTypes2 | EmailField       | Email Field        | Email             |                    |
+      | AllDataTypes2 | MultiSelectField | Multi Select Field | MultiSelectList   | regionalCentreEnum |
+      | AllDataTypes2 | CaseHistory      | History            | CaseHistoryViewer |                    |
+    Given the definition sheet 'FixedLists' looks like this
+      | ID                 | ListElementCode      | ListElement       |
+      | marritalStatusEnum | MARRIAGE             | Marriage          |
+      | marritalStatusEnum | CIVIL_PARTNERSHIP    | Civil Partnership |
+      | marritalStatusEnum | SINGLE	Single        |                   |
```
noronhaa: make it pretty

test/features/caseList.feature
```diff
@@ -5,6 +5,10 @@ Feature: Set of Scenarios testing the functionality of the Case List table
     Given I have logged in
 
   Scenario: A Case Reference column in case list results has hyphens every 4 digits
+    Given the definition sheet 'CaseEventToFields' looks like this
```
mario-paniccia: I think we should not mention or tie ourselves in any way to spreadsheets in our acceptance tests. We should be agnostic to the mechanism we use to define our definition. 
At the moment is in a spreadsheet, tomorrow it could be json based, or via the Admin Web UI only. But anyway this is not important here. 
I think we spoke about this before. 
It's very easy to refactor this in a more abstract language. And I strongly suggest we do.
Something like:
`Given the definition for 'event fields' looks like this` 

When we implement these steps, we will never interact with the spreadsheet directly. We will retrieve the definition from the database to validate what we declared
here is what is on the definition database. How it ended up there (via spreadsheet upload at the moment) we don't care, and should not be mentioned here.

test/features/caseList.feature
```diff
@@ -5,6 +5,10 @@ Feature: Set of Scenarios testing the functionality of the Case List table
...
+      | CaseTypeID    | CaseEventID   | CaseFieldID          | DisplayContext  |
```
mario-paniccia: I'd use the following formatting:
`| case type id| case event id| case field id| display context|`

test/features/caseList.feature
```diff
@@ -5,6 +5,10 @@ Feature: Set of Scenarios testing the functionality of the Case List table
     Given I have logged in
 
   Scenario: A Case Reference column in case list results has hyphens every 4 digits
+    Given the following definition for 'case fields'
```
mario-paniccia: not sure I understand the value of having this step in this particular scenario. The scenario is not testing anything related to an optional field

test/features/caseProgression.feature
```diff
@@ -1,13 +1,26 @@
-@caseProgression
+@functional @caseProgression
 Feature: Set of scenarios to test case creation and case progression
 
   Background:
+    Given the following definition for 'events'
+      | ID                 | Name                    | Description             | Display order | Pre condition sate | Post condition state |
+      | createCase         | Create a case           | Create a case           | 1             |                    | CaseCreated          |
```
mario-paniccia: do we need both Name and Description column? Also Display order perhaps is not much relevant in this case and could be removed?

test/features/caseView.feature
```diff
@@ -1,7 +1,21 @@
...
+      | Tab1             | Tab1              | 2                   | TextField1           | 1                   |                                  |
+      | Tab3             | Tab3              | 4                   | ConditionalText      | 3                   |                                  |
+      | Tab3             | Tab3              | 4                   | ConditionalFixedList | 3                   |                                  |
+      | Tab3             | Tab3              | 4                   | ConditionalYesNo     | 3                   |                                  |
+      | ConditionalTab1  | Conditional Tab 1 | 5                   | TextField5           | 1                   | ConditionalText="showmethemoney" |
+      | ConditionalTab2  | Conditional Tab 2 | 6                   | TextField6           | 1                   | ConditionalFixedList="TRUE"      |
+      | ConditionalTab3  | Conditional Tab 3 | 7                   | TextField7           | 1                   | ConditionalYesNo="Yes"           |
+
     Given I have logged in
```
mario-paniccia: I can see in the source file a step called:
`Given a case containing Tabs functionality exists`
I was thinking, is this step still needed now that we have the new step `Given the following definition for 'tab case fields'`? Aren't both steps verifying the same thing?

test/features/caseView_printButton.feature
```diff
@@ -0,0 +1,23 @@
+@functional @caseView
+Feature: Set of scenarios testing case view functionality
```
mario-paniccia: .. case view print button functionality...?

test/features/caseView_printButton.feature
```diff
@@ -0,0 +1,23 @@
...
+
+  Background:
+    Given I have logged in
+
+  Scenario: print button visible on case view when configured
+    Given the following definition for 'case tape'
+      | ID              | Name                 | PrintableDocumentsUrl                                                                     |
+      | AllDataTypes2   | All Field Data Types | http://localhost:4452/callback/jurisdictions/AUTOTEST1/case-types/AllDataTypes2/documents |
```
mario-paniccia: anything better we can think to avoid having the print url in the scenario?
For example we could just check whether there is any value (i.e. PrintableDocumentsUrl is not empty)

test/features/checkYourAnswers.feature
```diff
@@ -2,6 +2,12 @@
 Feature: Set of scenarios to test functionality of the Check Your Answers page
 
   Background:
+    Given the following definition for 'case fields in events'
+      | Event         | ID             | FieldType       | Display context | Page id         | Page label                    |
```
mario-paniccia: this step configuration doesn't seem relevant to me to the scenario we are testing. No where in the scenario we refer directly to the listed fields, or pages, or page label, or events. 
Either we remove it, or we make it more tailored to the scenario. 
For example we could show in the step that Show Your Answers is enabled in the definition for the field the scenario is using (adding a column 'check your answers' with values yes or no) 

ShowSummaryChangeOption
--

test/features/complexTypes.feature
```diff
@@ -1,7 +1,42 @@
 @functional @complexcollections @showandhide
-Feature: Set of scenarios testing show & hide functionality and conditional logic
```
mario-paniccia: Looking at the scenarios here and the definition step, I think this should be the feature file for complex type override per event. And we could have a separate feature file for complex types, which won't require the  `    Given the following definition for 'complex case fields overrides'` step

test/features/complexTypes.feature
```diff
@@ -1,7 +1,42 @@
...
+      | MySchool | Class.ClassMembers.Children.ChildFullName                         | MANDATORY       | Child full name (UPDATED) | Child hint (UPDATED) |                                                 |
+      | MySchool | Class.ClassMembers.Children.ChildGender                           | MANDATORY       |                           |                      |                                                 |
+      | MySchool | Class.ClassMembers.Children.ChildAddress.AddressLine1             | OPTIONAL        |                           |                      |                                                 |
+      | MySchool | Class.ClassMembers.Children.NeedsSupport                          | OPTIONAL        |                           |                      |                                                 |
+      | MySchool | Class.ClassMembers.Children.AutisticChildCaseNumber.CaseReference | OPTIONAL        |                           |                      |                                                 |
     Given I have logged in
 
   Scenario: limits the complex type using the EventToComplexTypes sheet
```
mario-paniccia: can you change `limits the complex type using the EventToComplexTypes sheet` to `limits the complex type using the Event to Complex Type override feature` please

test/features/caseList.feature
mario-paniccia: One more thing I've noticed in this PR, is that we added the new 'give the following definition step' to scenarios written in the old style. In my mind, we would have used this step only in scenarios written in the 'new style', i.e. the ones where all the fields mentioned in the 'given definition step' are then actually explicitly mentioned in the rest of the scenario. For example the scenario written in https://github.com/hmcts/ccd-case-management-web/pull/710/files. 
Maybe this is a good idea. The only concern I have is that it might result a bit confusing for who reads the scenario, because it doesn't understand why he has to pay attention to all the fields in the 'definition step' if they are then apparently not used in the scenario (but we know they are used under the hood).
I wanted to mention this to stimulate a discussion

test/features/search.feature
```diff
@@ -2,6 +2,36 @@
...
+      | NumberField      | Number Field       | Number            |                    |
+      | YesNoField       | Yes or No Field    | YesOrNo           |                    |
+      | CollectionField  | Collection Field   | Collection        | Text               |
+      | MarritalStatus   | Fixed List         | FixedList         | marritalStatusEnum |
+      | MoneyField       | Money Field        | MoneyGBP          |                    |
+      | DocumentField    | Document Field     | Document          |                    |
+      | EmailField       | Email Field        | Email             |                    |
+      | MultiSelectField | Multi Select Field | MultiSelectList   | regionalCentreEnum |
+      | CaseHistory      | History            | CaseHistoryViewer |                    |
+    Given the following definition for 'case enums'
```
mario-paniccia: what do you mean by  'case enums'?
shall we change this to:
 `Given the following definition for 'fixed lists'` ?

test/features/showandhide/collectionOfComplex.feature
```diff
@@ -2,27 +2,11 @@
...
-    Given the definition sheet 'EventToComplexTypes' looks like this
-      | ID	        | CaseEventID  | CaseFieldId | ListElementCode                          | FieldShowCondition                                                                                       |
-      | SchoolClass | createSchool | MySchool    | Name                                     |                                                                                                          |
-      | SchoolClass | createSchool | MySchool    | ProvidesAutisticChildrenSupport          |                                                                                                          |
-      | SchoolClass | createSchool | MySchool    | Class.ClassMembers.Children.IsAutistic   | MySchool.ProvidesAutisticChildrenSupport="Yes"                                                           |
-      | SchoolClass | createSchool | MySchool    | Class.ClassMembers.Children.NeedsSupport | MySchool.ProvidesAutisticChildrenSupport="Yes" AND MySchool.Class.ClassMembers.Children.IsAutistic="Yes" |
-
+    Given the following definition for 'case fields'
```
mario-paniccia: 👍

mario-paniccia: Can we rename this step:
`Given the following definition for 'complex case fields'` ?
Just makes it a bit more explicit the fact that these are complex fields. As it is, it's only understandable looking at the presence of the Path column

test/features/caseList.feature
noronhaa: > One more thing I've noticed in this PR, is that we added the new 'give the following definition step' to scenarios written purely in the old style. For example the ones in `checkYourAnswers.feature`.
> In my mind, we would have used this step only in scenarios written in the 'new style', i.e. the ones where all the fields mentioned in the 'given definition step' are then actually _explicitly_ mentioned in the rest of the scenario. For example the scenario written in https://github.com/hmcts/ccd-case-management-web/pull/710/files.
> Maybe this is a good idea and I can see the benefit it'll bring(makes sure the test is in sync with the definition). We can keep this.
> The only concern I have is that it might result a bit confusing for who reads the scenario and is not a dev, because he won't understand why he has to pay attention to all the fields in the 'definition step' if they are then _apparently_ not used in the scenario (but we know they are used under the hood). I was confused at the beginning too.
> I wanted to raise this to stimulate a discussion

I agree with this. no point just putting it in all features/scenarios for the sake of it. if it brings no value to the test/feature then we should leave it out

rafalkalita: will update

test/features/caseProgression.feature
rafalkalita: will change

test/features/caseView_printButton.feature
rafalkalita: not mine, but will update

test/features/caseView_printButton.feature
rafalkalita: 👍

test/features/checkYourAnswers.feature
rafalkalita: I think it is not relevant for the test. It is testing we can change the value and all field are enabled. It would be relevant if we were checking visibility of the field when using `ShowSummaryChangeOption`.

test/features/complexTypes.feature
rafalkalita: 👍

test/features/complexTypes.feature
rafalkalita: Will change to:
Scenario: filters the complex type fields using the Event to Complex Type override feature
Scenario: overrides the complex type data using the Event to Complex Type override feature

test/features/search.feature
rafalkalita: 👍

test/features/showandhide/collectionOfComplex.feature
rafalkalita: 👍

PR #732, opened 2019-06-28
RDM-5031 - displaying thrown alert as it is instead of clearing based on search result error.

smathangi:
https://tools.hmcts.net/jira/browse/RDM-5031

### Change description ###
As per Andrew, this condition was added during save draft story.  But it seems draft flow is working fine without this condition. 
https://tools.hmcts.net/jira/browse/RDM-2302

```
[ ] Yes
[ ] No
```


src/app/workbasket/workbasket.component.ts
```diff
@@ -93,8 +93,9 @@ export class WorkbasketComponent implements OnInit {
           this.caseState = filter.caseState;
           this.page = filter.page;
           this.paginationMetadata = results[1];
-          // Clearing the errors is only on the assumption this is the only place we display errors on case list page
-          this.resultView.result_error ? this.alertService.warning(this.resultView.result_error) : this.alertService.clear();
+          if (this.resultView.result_error) {
```
mario-paniccia: if we remove the clear, we'll see the error banner as long as we stay on the workbasket page isn't it? Whilst before if you changed filters and did a search, the error banner was cleared. I mean, there's a change in behaviour here.

smathangi: as discussed in the slack, it's clearing the banner upon a click on apply button with changed filters.

mario-paniccia: 👍

PR #756, opened 2019-07-24
RDM-4648- Functional tests for dynamicists

mohammedanas:
https://tools.hmcts.net/jira/browse/RDM-4648
Scenario1:
Given: I have a Dynamic fixed list field defined on Page 2 during create event
When: I return the Dynamic fixed list values on About to start callback
Then: I should see the Dynamic fixed list field defined on Page 2 populated with the fields returned during the About to start callback in the order defined.
Scenario2:
Given: I have Dynamic fixed list field defined on Page2 during create event
When: I return the Dynamic fixed list values on midevent callback after Page1
Then: I should see the Dynamic fixed list field defined on Page 2 populated with the fields returned during the midevent callback in the order definied.

Scenario3:
Given: I have Dynamic fixed list field defined on Page2 during create event
When: I return the Dynamic fixed list values on About to start callback with default selection
Then: I should see the Dynamic fixed list field defined on Page 2 populated with the fields returned during the About to start callback in the order defined and default selection.

```
[ ] Yes
[ ] No
```


test/stepDefinitions/baseSteps.js
```diff
@@ -12,6 +12,13 @@ module.exports = {
     await createCaseStartPage.clickStartButton();
   },
 
+  navigateToCreateCaseMultiPage: async function(){
```
noronhaa: lets remove this as it has a hardcoded case type and is not very reusable. instead use navigateToCreateCasePage and pre load the data through a (new?) previous step implemented in definitionSteps.js 
either use existing step `Given a case type with multiple pages exist`
or new one with same implementation `Given a case type with multiple pages containing a dynamic fixed list exists`

test/features/fieldTypes.feature
```diff
@@ -69,6 +69,59 @@ Feature: Set of scenarios to check we can read and write to all field data types
       | dataType  |
       | Complex   |
```
noronhaa: too much copy and pasting perhaps - some of these scenarios are for mid event callback i thought however all the titles say they are all for AboutToStart callback

test/features/fieldTypes.feature
```diff
@@ -69,6 +69,59 @@ Feature: Set of scenarios to check we can read and write to all field data types
...
+    |--Select a value--|
+    | List 1|
+    | List 2|
+    | List 3|
+    | List 4|
+    | List 5|
+    | List 6|
+    | List 7|
+
+  Scenario: Dynamic fixed list is populated with values from AboutToStart Callback
```
noronhaa: looks like a duplicated scenario?

test/stepDefinitions/create-caseSteps.js
```diff
@@ -174,6 +174,18 @@ defineSupportCode(function ({ Given, When, Then, And}) {
     await baseSteps.navigateToCreateCasePage();
   });
 
+  When(/^I navigate to multiple pages case type form pages$/, async function (){
+    await baseSteps.navigateToCreateCaseMultiPage();
+    await new CreateCaseWizardPage().clickContinueButton();
+  })
+
+  When(/^I create a case with multiple pages$/, async function (){
```
noronhaa: see if we can use the existing step `And I create the case` instead of this, should work then delete this step

PR #762, opened 2019-07-30
Rdm 4617

mohammedanas:
https://tools.hmcts.net/jira/browse/RDM-4617

Currently the extracted case view component has fixed navigation routes (to the surrounding consumer these are routes outside of viewer routing) as part of its contract on events such as:

when redirected to case view after a case created, and the user has no READ access routes to
/list/case
(case.resolver.ts)
when on view draft page and when deleting the draft is triggered cancelling and discarding the draft routes to
/list/case
(case.viewer.component.ts)
when on view draft page and when resuming the draft is triggered cancelling and discarding the draft routes to

/create/case/:jid/:ctid/:eid
(case.viewer.component.ts)

The navigation should be a concern of the consuming component thus event emitters should be added that would emit contextual events so that consumers could decide what action to perform upon receiving them (e.g. which page to navigate to).

```
[ ] Yes
[X] No
```


src/app/core/utils/navigation-listener.service.spec.ts
```diff
@@ -0,0 +1,62 @@
...
+    mockRouter.navigate.and.returnValue(Promise.resolve(true));
+    navigationNotifier = new NavigationNotifierService();
+    navigationListenerService = new NavigationListenerService(alertService, navigationNotifier, mockRouter);
+  });
+
+  it('test DRAFT_DELETED navigation subscription', () => {
+    let DRAFT_DELETED = NavigationOrigin.DRAFT_DELETED;
+    navigationNotifier.announceNavigation({ action: DRAFT_DELETED });
+    navigationListenerService.init();
+    expect(mockRouter.navigate).toHaveBeenCalledWith(['list/case']);
```
andrewfolga: We also need to notify as follows:
```
this.alertService.setPreserveAlerts(true);
this.alertService.success(`The draft has been successfully deleted`);
```

src/app/core/utils/navigation-listener.service.spec.ts
```diff
@@ -0,0 +1,62 @@
...
+  });
+
+  it('test ERROR_DELETING_DRAFT navigation subscription', () => {
```
andrewfolga: Need to also handler error:
```
  private handleError(error: HttpError, trigger: CaseViewTrigger) {
    if (error.status !== 401 && error.status !== 403) {
      this.error = error;
      console.log('error during triggering event:', trigger.id);
      console.log(error);
      this.callbackErrorsSubject.next(this.error);
    }
  }
```

src/app/core/utils/navigation-listener.service.spec.ts
```diff
@@ -0,0 +1,62 @@
...
+      navParam.ctid,
+      navParam.etid], { queryParams: navParam.theQueryParams });
+  });
+
+  it('test EVENT_TRIGGERED navigation subscription', () => {
+    let EVENT_TRIGGERED = NavigationOrigin.EVENT_TRIGGERED;
+    let navParam = { action: EVENT_TRIGGERED, relativeTo: 'relativeTo', etid: 'e001', theQueryParams: 'id=1001' };
+    navigationNotifier.announceNavigation(navParam);
+    navigationListenerService.init();
+    expect(mockRouter.navigate).toHaveBeenCalledWith(['trigger',
```
andrewfolga: Need to also handler error:

  private handleError(error: HttpError, trigger: CaseViewTrigger) {
    if (error.status !== 401 && error.status !== 403) {
      this.error = error;
      console.log('error during triggering event:', trigger.id);
      console.log(error);
      this.callbackErrorsSubject.next(this.error);
    }
  }

src/app/core/utils/navigation-listener.service.spec.ts
```diff
@@ -0,0 +1,62 @@
...
+    expect(mockRouter.navigate).toHaveBeenCalledWith(['/list/case']);
+  });
+
+  it('test DRAFT_RESUMED navigation subscription', () => {
+
+    let DRAFT_RESUMED = NavigationOrigin.DRAFT_RESUMED;
+    let navParam = { action: DRAFT_RESUMED, jid: 'j001', ctid: 'c001', etid: 'e001', theQueryParams: 'id=1001' };
+    navigationNotifier.announceNavigation(navParam);
+    navigationListenerService.init();
+    expect(mockRouter.navigate).toHaveBeenCalledWith(['create/case',
```
andrewfolga: Need to also handler error:
```
  private handleError(error: HttpError, trigger: CaseViewTrigger) {
    if (error.status !== 401 && error.status !== 403) {
      this.error = error;
      console.log('error during triggering event:', trigger.id);
      console.log(error);
      this.callbackErrorsSubject.next(this.error);
    }
  }
```

src/app/core/utils/navigation-listener.service.spec.ts
```diff
@@ -0,0 +1,62 @@
...
+    navigationNotifier.announceNavigation({ action: ERROR_DELETING_DRAFT });
+    navigationListenerService.init();
+    expect(mockRouter.navigate).toHaveBeenCalledWith(['list/case']);
+  });
+
+  it('test NO_READ_ACCESS_REDIRECTION navigation subscription', () => {
+    let NO_READ_ACCESS_REDIRECTION = NavigationOrigin.NO_READ_ACCESS_REDIRECTION;
+    navigationNotifier.announceNavigation({ action: NO_READ_ACCESS_REDIRECTION });
+    navigationListenerService.init();
+    expect(mockRouter.navigate).toHaveBeenCalledWith(['/list/case']);
```
andrewfolga: We need to also make sure we announce success:
```
.then(() => this.alertService.success(CaseResolver.CASE_CREATED_MSG));
```

src/app/core/utils/navigation-listener.service.ts
```diff
@@ -0,0 +1,59 @@
...
+          });
+        case NavigationOrigin.EVENT_TRIGGERED:
+          return this.router.navigate(['trigger', navigation.etid], {
+            queryParams: navigation.theQueryParams,
+            relativeTo: navigation.relativeTo
+          }).catch(error => {
+            this.handleError(error, navigation.etid);
+          });
+        case NavigationOrigin.NO_READ_ACCESS_REDIRECTION:
+          return this.router.navigate((['/list/case']));
```
andrewfolga: We need to alert successful creation of case here.

src/app/core/utils/navigation-listener.service.ts
```diff
@@ -0,0 +1,59 @@
...
+  callbackErrorsSubject: Subject<any> = new Subject();
+  navigationSubscription: Subscription;
+
+  constructor(
+    private alertService: AlertService,
+    private navigationNotifier: NavigationNotifierService,
+    private router: Router) {
+  }
+
+  public init() {
```
andrewfolga: Would it be possible to have an `ngOnInit()` instead of `init()` and whoever takes it in Module class would subscribe automatically? Maybe it is not a best option here though.

src/app/core/utils/navigation-listener.service.ts
tzarsmango: 👍

src/app/core/utils/navigation-listener.service.ts
tzarsmango: This is not a component but a service class, so I suspect ngOnInit would fire even if we implemented it. We call init method implicitly in the `CoreComponent's ngOnInit`

src/app/core/utils/navigation-listener.service.spec.ts
tzarsmango: done

src/app/core/utils/navigation-listener.service.spec.ts
tzarsmango: not sure if this needs error handling just a redirect to main page

src/app/core/utils/navigation-listener.service.spec.ts
tzarsmango: ok

src/app/core/utils/navigation-listener.service.spec.ts
tzarsmango: done

src/app/core/utils/navigation-listener.service.spec.ts
tzarsmango: done :+1

src/app/core/utils/navigation-listener.service.ts
andrewfolga: My bad. Good point

src/app/core/utils/navigation-listener.service.spec.ts
andrewfolga: Yes correct.

src/app/core/utils/navigation-listener.service.spec.ts
andrewfolga: Yes correct.

src/app/core/utils/navigation-listener.service.spec.ts
andrewfolga: Yes correct.

PR #775, opened 2019-08-13
RDM-4090 - hide sensitive app config

gbenadikar:
https://tools.hmcts.net/jira/browse/RDM-4090

### Change description ###
Draft PR to hide sensitive configurations

```
[ ] Yes
[x] No
```


set-env.ts
```diff
@@ -0,0 +1,40 @@
+import { writeFile } from 'fs';
+
+const environmentFilePath = `./src/environments/environment.ts`;
+const environmentFile = `
+export const environment = {
+  production: ${process.env['PRODUCTION'] || false},
```
mario-paniccia: this info is duplicated in `const CONFIG = {...}` in `server.ts`. We would have to update in 2 places now when we add some new config. Is it possible to remove somehow this duplication?

src/environments/environment.ts
```diff
@@ -1,9 +1,28 @@
-// The file contents for the current environment will overwrite these during build.
-// The build system defaults to the dev environment which uses `environment.ts`, but if you do
-// `ng build --env=prod` then `environment.prod.ts` will be used instead.
```
mario-paniccia: according to this, does the Prod pipeline expects environment.prod.ts to be an existing file, which we are now removing?

package.json
```diff
@@ -31,11 +32,11 @@
     "lint": "ng lint ccd-case-management-web",
     "e2e": "ng e2e",
     "start:fr": "ng serve --aot --i18nFile=src/locale/messages.fr.xlf --i18nFormat=xlf --locale=fr",
-    "build:fr": "ng build --prod --i18nFile=src/locale/messages.fr.xlf --i18nFormat=xlf --locale=fr",
+    "build:fr": "npm run config && ng build --prod --i18nFile=src/locale/messages.fr.xlf --i18nFormat=xlf --locale=fr",
```
mario-paniccia: rather than changing build:fr and build:client-and-server-bundles which both delegate to build, can't we change only `build`? I guess we want to call `npm run config` on every type of build?

package.json
```diff
@@ -13,6 +13,7 @@
   },
   "scripts": {
     "ng": "ng",
+    "config": "ts-node set-env.ts",
     "postinstall": "webdriver-manager update --versions.chrome 2.40",
     "build": "ng build",
```
mario-paniccia: do we nee to change this too?

src/environments/environment.prod.ts
```diff
@@ -1,4 +1,28 @@
+
 export const environment = {
   production: true,
-  configUrl: '/config'
+  login_url: 'http://localhost:3501/login',
```
mario-paniccia: do we need this content here? how is this one different than the one in set-env.ts? thinking if we can remove this duplication with set-env.ts

gbenadikar: yes can be left empty, will do

set-env.ts
gbenadikar: removed the duplication

src/environments/environment.ts
gbenadikar: reverted environment.prod.ts. The script set-env.ts will now populate environment.prod.ts on build

package.json
gbenadikar: we will need to run config script wherever we use the environment as 'prod'

PR #777, opened 2019-08-20
updated Chart

akdasari:

```
[ ] Yes
[x ] No
```

### DIFF
--- a/charts/ccd-case-management-web/values.yaml
+++ b/charts/ccd-case-management-web/values.yaml
@@ -1,14 +1,14 @@
 nodejs:
-  image: hmctspublic.azurecr.io/ccd/case-management-web:latest
-  applicationPort: 3451
-  cpuLimits: 1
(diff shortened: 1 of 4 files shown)


charts/ccd-case-management-web/values.yaml
```diff
@@ -1,17 +1,14 @@
 nodejs:
-  image: hmctspublic.azurecr.io/ccd/case-management-web:latest
-  applicationPort: 3451
-  memoryRequests: 256Mi
```
mario-paniccia: how come memoryRequests, cpuRequests, memoryLimits have been removed?

akdasari: lot of things are moved into base images like java or nodes, hence no need in sub charts

mario-paniccia: 👍

PR #827, opened 2019-11-12
RDM-5875 Update ccpay-web-component dependency

danlysiak:
https://tools.hmcts.net/jira/browse/RDM-5875

### Change description ###
Update the ccpay-web-component dependency

```
[ ] Yes
[x] No
```


danlysiak:
Note latest ccpay-web-component also requires `file-saver` and `xlsx` (confirmed with reporter of RDM-5875).

package.json
```diff
@@ -75,6 +75,7 @@
     "cucumberjs-allure-reporter": "1.0.3",
     "expect": "^23.0.0",
     "express": "^4.16.2",
+    "file-saver": "^2.0.2",
```
smathangi: is it required for pay-web-component?

danlysiak: Yep, the latest version requires both `file-saver` and `xlsx` - fails to compile without.

hemantt: Isnt it an overhead for us to maintain those standalone dependencies versions? @danlysiak @smathangi

smathangi: Agree @hemantt  ideally those should part of pay web-component dependency itself.

package.json
```diff
@@ -58,12 +58,12 @@
     "@angular/router": "~7.0.3",
     "@angular/upgrade": "~7.0.3",
     "@hmcts/ccd-case-ui-toolkit": "2.62.21-RDM-4335-prerelease-2",
-    "@hmcts/ccpay-web-component": "~2.0.50",
+    "@hmcts/ccpay-web-component": "^3.0.0-beta15",
```
mario-paniccia: I've found some references to `ccpay-web-component` in the toolkit as well. A bit old versions thought, so not sure if those are a leftover or still used. Can you please check whether we need to update there as well?

package.json
```diff
@@ -58,12 +58,12 @@
     "@angular/router": "~7.0.3",
     "@angular/upgrade": "~7.0.3",
     "@hmcts/ccd-case-ui-toolkit": "2.62.21-RDM-4335-prerelease-2",
-    "@hmcts/ccpay-web-component": "~2.0.50",
```
mario-paniccia: there's a reference to `ccpay-web-component` also in toolkit. Can you double check whether we have to update also there?  
https://github.com/hmcts/ccd-case-ui-toolkit/blob/c10b9b5bf98754e9f9c938bafc40eaef05ae42f3/package.json#L80

danlysiak: Checked with pay team and @hemantt - shouldn't be required.

package.json
danlysiak: 👍 I was advised yesterday that there were issues trying to incorporate into the component itself, but requested they look to sort in future versions so we no longer need to maintain them.

PR #836, opened 2019-11-21
RDM-6635 - pay bulkscan integration

smathangi:
https://tools.hmcts.net/jira/browse/RDM-6635


charts/ccd-case-management-web/values.aat.template.yaml
```diff
@@ -13,3 +13,4 @@ nodejs:
     PRINT_SERVICE_URL_REMOTE: https://return-case-doc-ccd.nonprod.platform.hmcts.net
     CCD_ACTIVITY_URL: https://gateway-ccd.aat.platform.hmcts.net/activity
     PAYMENTS_URL: https://gateway-ccd.aat.platform.hmcts.net/payments
+    PAY_BULKSCAN_URL: https://gateway-ccd.aat.platform.hmcts.net/pay-bulkscan
```
andrewfolga: Is this only needed for aat and preview?

smathangi: Yes, ideally for preview only. When frontend apps move to AKS in the next PI we've to refactor / cleanup these files and override if necessary in the flux config.

PR #845, opened 2019-12-13
RDM-6719: Add Banner Feature to CCD UI

kiran-yenigala-hmcts:

https://tools.hmcts.net/jira/browse/RDM-6719

### Change description ###

Show Banner on CCD UI for each Jurisdiction based on the configuration

```
[ ] Yes
[X ] No
```


src/app/cases/creator/case-creator.component.spec.ts
```diff
@@ -58,7 +58,11 @@ const JURISDICTION_1: Jurisdiction = {
   id: 'J1',
   name: 'Jurisdiction 1',
   description: '',
-  caseTypes: CASE_TYPES_1
+  caseTypes: CASE_TYPES_1,
```
rafalkalita: We do have a Banner object. Why not to compose it in here?

PR #850, opened 2019-12-18
RDM-4090 - hide sensitive app config - version 2

rafalkalita:
https://tools.hmcts.net/jira/browse/RDM-4090

### Change description ###
A simpler version which I am not sure if will work.

```
[ ] Yes
[ X] No
```


server.ts
```diff
@@ -79,9 +80,11 @@ const healthConfig = {
 healthcheck.addTo(appHealth, healthConfig);
 app.use(appHealth);
 
-app.get('/config', (req, res) => {
-  res.status(200).json(CONFIG);
-});
+if (environment.production !== true) {
```
MSancaktutar: Didn't look into the code yet, but at a first glance, will this work on ITHC env? If not, ITHC will fail.

MSancaktutar: And, isn't there any use on prod of this endpoint? Again, asking at a superficial glance..

rafalkalita: It depends how do they deploy it to ITHC env. If they deploy in the prod mode enabled then /conf should be disabled. The card is about restricting the config endpoint for prod env.

PR #859, opened 2020-01-15
RDM-6398: add UI functional test for Case View displaying of errors

mattnayler:

https://tools.hmcts.net/jira/browse/RDM-6398

### Change description ###

Adds the following new test elements:

* features:
  - caseViewCallbackMessages.feature
* pageObjects:
  - ccd-components/callbackErrorsComponent.js
* stepDefinitions:
  - callbackErrorsSteps.js
* definitionsFiles:
  - fe-automation-definition-v30-RDM-6398_caseview_callback_errors.xlsx

Updates:

* pageObjects:
  - caseDetailsPage.js - alert bar, error summary, Go button
* stepDefinitions:
  - caseDetailsSteps.js - alert bar, error summary, Go button
  - definitionFileSteps.js - new CaseView Callback case type

NB: dependent on new stubs: hmcts/ccd-test-stubs-service#35

```
[ ] Yes
[x] No
```


test/features/caseViewCallbackMessages.feature
```diff
@@ -0,0 +1,90 @@
...
+      | triggerCallbackErrorMultiple     | Trigger callback errors       | CaseCreated        | AfterEvent           |
+      | triggerCallbackErrorsAndWarnings | Trigger callback errs + warns | CaseCreated        | AfterEvent           |
+    Given the following definition for 'states'
+      | ID          | Name         |
+      | CaseCreated | Case created |
+      | AfterEvent  | After Event  |
+
+    Given I have logged in
+    And a case with CaseView Callback functionality exists
+
```
mario-paniccia: This PR looks very good.
I only have one important comment. According to the testing pyramid principle, we should rely as little as possible on functional tests. Ideally we should strictly test only 'the basics' of the functionality with functional tests, and then rely on faster and less expensive (in terms of maintainability) integration test and unit tests to test all the variations and corner cases of the feature. This also prevents our func tests suite to take a large amount of time to complete.
Because of the above, I would strongly suggest we only leave the following func tests and remove the others:
`Callback errors and warnings displayed when start event triggers multiple errors and warnings`
` Starting event twice will ignore callback warnings generated when start event triggers multiple warnings`

PR #888, opened 2020-02-21
RDM-2508-Rm-nsp

nathan-clark:
```
[ ] Yes
[x] No
```


package.json
```diff
@@ -74,8 +73,6 @@
     "cucumber-html-reporter": "^4.0.5",
     "cucumber-pretty": "1.5.0",
     "expect": "^23.0.0",
-    "xlsx": "^0.15.1",
-    "file-saver": "^2.0.2",
```
hemantt: @nathan-clark these xlsx and file-saver dependencies are needed for pay lib. Please could you add them back.

nathan-clark: I just removed nsp, which removed these direct dependencies?!?

PR #913, opened 2020-03-16
RDM-7932 - demo print url update

smathangi:

https://tools.hmcts.net/jira/browse/RDM-7932

### Change description ###

CCD Print URL doesn't work correctly on Demo

```
[ ] Yes
[ x] No
```

### DIFF
--- a/infrastructure/demo.tfvars
+++ b/infrastructure/demo.tfvars
@@ -1,7 +1,7 @@
 idam_authentication_web_url = "https://idam-web-public.demo.platform.hmcts.net/"
 ccd_gateway_url = "https://gateway-ccd.demo.platform.hmcts.net"
 document_management_url = "^https?://(?:api-gateway\\.preprod\\.dm\\.reform\\.hmcts\\.net|dm-store-demo\\.service\\.core-compute-demo\\.internal(?::\\d+)?)"
-ccd_print_service_url = "http://return-case-doc.ccd.demo.platform.hmcts.net"
+ccd_print_service_url = "https://return-case-doc.ccd.demo.platform.hmcts.net"
 activity_enabled = "true"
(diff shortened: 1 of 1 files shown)


infrastructure/demo.tfvars
```diff
@@ -1,7 +1,7 @@
 idam_authentication_web_url = "https://idam-web-public.demo.platform.hmcts.net/"
 ccd_gateway_url = "https://gateway-ccd.demo.platform.hmcts.net"
 document_management_url = "^https?://(?:api-gateway\\.preprod\\.dm\\.reform\\.hmcts\\.net|dm-store-demo\\.service\\.core-compute-demo\\.internal(?::\\d+)?)"
-ccd_print_service_url = "http://return-case-doc.ccd.demo.platform.hmcts.net"
+ccd_print_service_url = "https://return-case-doc.ccd.demo.platform.hmcts.net"
```
mario-paniccia: @MSancaktutar I can see from the history you changed this url from `https` to `http`. With a commit message:
"fix downstream url and push to demo". Any reason for that change? Just making sure reverting your change doesn't create other issues.

MSancaktutar: Thanks for confirming @mario-paniccia. I can't recall now, but most most probabl I should have done it because it would work with http. If the http comms is (no more) available to print service, obviously we should now change it to https, which is what's already done on this PR and I'm happy with it.

smathangi: As per code : it's not actually print service url; it's PRINT_SERVICE_URL_REMOTE which need to be inline with ccd-data-store config https://github.com/hmcts/ccd-data-store-api/blob/master/infrastructure/demo.tfvars#L3 to be able to route via api-gateway proxy

mario-paniccia: 👍

PR #914, opened 2020-03-17
Rdm-7924: Handle no user profile in the front end

kiran-yenigala-hmcts:
https://tools.hmcts.net/jira/browse/RDM-7924

### Change description ###

Handle profile not having workbasket defaults

```
[ ] Yes
[ X] No
```

### DIFF
--- a/src/app/core/core.component.ts
+++ b/src/app/core/core.component.ts
@@ -51,7 +51,7 @@ export class CoreComponent implements OnInit, OnDestroy {
         this.selectedJurisdictionName = selectedJurisdiction.name;
       }
     );
-    if (this.profile.default.workbasket.jurisdiction_id) {
+    if (this.profile.default.workbasket && this.profile.default.workbasket.jurisdiction_id) {
       this.jurisdictionService.announceSelectedJurisdiction(this.profile.jurisdictions.find(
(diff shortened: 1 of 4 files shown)


src/app/core/core.component.ts
```diff
@@ -51,7 +51,7 @@ export class CoreComponent implements OnInit, OnDestroy {
         this.selectedJurisdictionName = selectedJurisdiction.name;
       }
     );
-    if (this.profile.default.workbasket.jurisdiction_id) {
+    if (this.profile.default.workbasket && this.profile.default.workbasket.jurisdiction_id) {
```
mario-paniccia: wondering why this is needed. Has the json response coming from the backend changed in the case where no defaults are present compared to what we had before for this case?

kiran-yenigala-hmcts: https://github.com/hmcts/ccd-data-store-api/pull/819/files/edb6ee43c9353ebf9c78cf44c185edf0e606ca89#diff-8130758e1cdebc332c7d2e7a41b4962aR70

Definition store throwing exception when no workbasket defaults were found... and handled in data store and not sending WorkbasketDefault in response now....

PR #933, opened 2020-04-22
RDM-7854 ManagementWeb build failure

RebeccaBaker:
https://tools.hmcts.net/jira/browse/RDM-7854

### Change description ###
ManagementWeb build failures intermittent 

```
[ ] Yes
[X] No
```

### DIFF
--- a/test/stepDefinitions/validationSteps.js
+++ b/test/stepDefinitions/validationSteps.js
@@ -145,6 +145,7 @@ defineSupportCode(function ({ Given, When, Then, Before, After }) {
   });
 
   When(/^I enter '(.*)' into the '(.*)' field$/, async function (value, fieldType) {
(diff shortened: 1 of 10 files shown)


test/stepDefinitions/validationSteps.js
```diff
@@ -145,6 +145,7 @@ defineSupportCode(function ({ Given, When, Then, Before, After }) {
   });
 
   When(/^I enter '(.*)' into the '(.*)' field$/, async function (value, fieldType) {
+    browser.wait(ExpectedConditions.presenceOf(element(by.css('ccd-write-money-gbp-field'))), 5000);
```
danlysiak, line 148: This would only work for case types which have a money field on their first page. But since it passes with the  current test data and case-management-web isn't expecting more changes, a more generic solution probably not worth looking into.

PR #1026, opened 2021-02-22
RDM-11245: Add logging for page validation

rajaram-nagarajan:
### JIRA link  ###

https://tools.hmcts.net/jira/browse/RDM-11245

### Change description ###

Added logging for page validation

```
[ ] Yes
[x] No
```


server.ts
```diff
@@ -51,7 +51,9 @@ const CONFIG = {
   'appInsights_enabled': process.env['APPINSIGHTS_ENABLED'] || 'false',
   'appInsights_roleName': process.env['APPINSIGHTS_ROLE'] || 'ccd-case-management-web',
   'shutter_redirect_url': process.env['SHUTTER_REDIRECT_URL'] || '',
-  'shutter_redirect_wait': parseInt(process.env['SHUTTER_REDIRECT_WAIT'], 10) || 10
+  'shutter_redirect_wait': parseInt(process.env['SHUTTER_REDIRECT_WAIT'], 10) || 10,
+  'logging_level': process.env['LOGGING_LEVEL'] || 'Off',
```
MSancaktutar, line 55: `Off` doesn't read like a logging level. Perhaps change var name?

server.ts
```diff
@@ -51,7 +51,9 @@ const CONFIG = {
...
+  'logging_case_field_list': process.env['LOGGING_CASE_FIELD_LIST'] || ''
```
MSancaktutar, line 56: would we need this in fact?

rajaram-nagarajan: Yes, we need this to avoid build issues

server.ts
rajaram-nagarajan: Removed 'Off' from the logging level.

server.ts
rajaram-nagarajan: Removed logging

MSancaktutar (approved):
LGTM

PR #1132, opened 2021-11-23
CCD-2209: Update fontconfig version in Dockerfile

paulridout:
CCD-2209 (https://tools.hmcts.net/jira/browse/CCD-2209)

### Change description ###
- Updated fontconfig entry in Dockerfile to reference version 2.13.1-r4 instead of 2.13.1-r2.  Build log shows that Docker build stage uses version 3.14 of Alpine linux which contains version 2.13.1-r4 of fontconfig rather than r2.

```
[ ] Yes
[x] No
```


kaanaktas (approved):
lgtm

